PR #3, opened 2023-09-02
Exclude large QASM test files from language analysis + formatting clean-up

chemix-lunacy:
Some basic sanity formatting clean-up along with basic things like CI and language detection fix.


chemix-lunacy:
Merging due to no functional changes, repo config refinement (which I need to see if it works) and we want some form of CI in.

PR #5, opened 2023-09-11
Add skeleton for verification engine

chemix-lunacy:
Very little here, but a template for other people to extend later.

### DIFF
--- a/src/QAT/qat/purr/backends/verification.py
+++ b/src/QAT/qat/purr/backends/verification.py
@@ -1,0 +1,113 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Oxford Quantum Circuits Ltd
+
+from abc import ABC, abstractmethod
+from typing import List, Optional
+
+from qat.purr.backends.live import LiveHardwareModel
+from qat.purr.compiler.builders import InstructionBuilder
+from qat.purr.compiler.emitter import QatFile
+from qat.purr.compiler.execution import QuantumExecutionEngine
+from qat.purr.compiler.instructions import Instruction
+
+
+class QPUVersion:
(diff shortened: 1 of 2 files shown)


src/QAT/qat/purr/backends/verification.py
```diff
@@ -0,0 +1,67 @@
...
+
+    if qpu_type.make == Lucy.__name__:
+        # TODO: Should have an apply_setup_to_hardware in live which people can use to build our own architecture.
+        model = VerificationModel(qpu_type, LucyVerificationEngine)
+        raise NotImplementedError("No lucy-specific model created yet.")
+
+    return None
+
+
+class VerificationEngine(QuantumExecutionEngine, ABC):
```
jfriel-oqc, line 60: Think it's worth adding an abc verify_instructions on this and the model? I like the override api, but also having an instructions verifier wrapped in a results object would be clean for anyone looking to submit instructions over the cloud

jfriel-oqc (approved):

src/QAT/qat/purr/backends/verification.py
chemix-lunacy: You mean a free-standing method that just takes in instructions? I mean that'd be fine, but people will need to know that you _have_ to have already passed the verifying instructions through the full pipeline before calling it. You can't just parse them and verify.

Also they won't be in results - it'll just have a composite exception thrown with all the issues with it, unless you meant something else?

chemix-lunacy: I've added a free-standing method for people to use, but these verifiers will never return a result - they will just exception with everything wrong in the messages. It's a verifier for correctness, not inefficiencies.

PR #6, opened 2023-09-19
Added timing metrics to CompilerMetrics object. Bump Compiler Config …

harry-oqc:
I have added the following metrics to the CompilationMetrics object to track the timings of the compilation/execution process.

ExecutionDuration
OptimizationDuration
ParseDuration
EngineCallDuration
TotalDuration
As metric enums have been changed, this has necessitated a version change to the compiler config


src/QAT/qat/purr/compiler/execution.py
```diff
@@ -219,15 +224,20 @@ def _generate_repeat_batches(self, repeats):
         )
         return ([batch_limit] * list_expansion_ratio) + [repeats % batch_limit]
 
-    def execute(self, instructions):
+    def execute(self, instructions) -> Dict[str, Any]:
```
chemix-lunacy, line 227: This can just be moved up to `InstructionExecutionEngine`, it no longer needs to be abstract.

src/QAT/qat/qat.py
```diff
@@ -74,8 +75,13 @@ def execute_with_metrics(
...
 
-    frontend: LanguageFrontend = fetch_frontend(path_or_str)
-    return _execute_with_metrics(frontend, path_or_str, hardware, compiler_config)
+    total_execution_timer = log_duration("Program execution completed. Total runtime took {} seconds.")
+    with total_execution_timer:
+        frontend: LanguageFrontend = fetch_frontend(path_or_str)
+        results, metrics = _execute_with_metrics(frontend, path_or_str, hardware, compiler_config)
+    metrics.record_metric(MetricsType.TotalDuration, total_execution_timer.duration)
+
+    return results, metrics.as_dict()
```
chemix-lunacy, line 84: This should all be moved to `_execute_with_metrics`. As a side note, you changed the return value of every other entry-point API - `execute_qir_x` etc.

src/QAT/qat/purr/compiler/runtime.py
```diff
@@ -256,7 +255,11 @@ def execute_instructions(
     active_runtime = get_runtime(hardware)
 
     active_runtime.run_quantum_executable(executable_blocks)
+
+    results, engine_call_duration_metrics = active_runtime.execute(instructions, results_format, repeats)
```
chemix-lunacy, line 259: Find where `active_runtime.compilation_metrics` is used and remove it from the runtime, instead returning it in the new metrics object. It was always going to be moved anyway, since runtimes persist beyond a single running instance. 

You can then turn all of this into just:

```
return active_runtime.execute(...)
```

src/tests/test_config.py
```diff
@@ -157,12 +157,12 @@ def test_config_deserialization_raises_error(self):
...
 
     def test_json_version_compatibility(self):
-        serialised_data = _get_contents("serialised_default_compiler_config_v1.json")
+        serialised_data = _get_contents("serialised_default_compiler_config_v2.json")
         deserialised_conf = CompilerConfig.create_from_json(serialised_data)
         assert deserialised_conf.metrics == MetricsType.Default
         assert deserialised_conf.results_format == QuantumResultsFormat()
 
-        serialised_data = _get_contents("serialised_full_compiler_config_v1.json")
+        serialised_data = _get_contents("serialised_full_compiler_config_v2.json")
```
chemix-lunacy, line 165: You actually need to add new tests, because you still want to know if v1 serializes correctly.

chemix-lunacy (changes requested):
Looks good! Need to fix the issues raised but otherwise happy to approve.

src/tests/test_config.py
chemix-lunacy: Well if that's the case then you're going to need to do some work to properly map them from v1 to v2 as a part of this PR and make sure you can transmute v1 to v2, since nothing you've done here is un-transmutable. 

I can add this to the PR if you don't have time, but this can't be put in with the capability to transmute from v1 to v2.

harry-oqc: Yes this makes sense - I have very low capacity at the moment but can try to get to this some point in middle of next week if you do not have time to add this before :)

src/QAT/qat/purr/compiler/execution.py
```diff
@@ -68,7 +73,7 @@ def run_calibrations(self, qubits_to_calibrate=None):
         pass
 
     @abc.abstractmethod
-    def execute(self, instructions: List[Instruction]) -> Dict[str, Any]:
+    def execute_with_metrics(self, instructions: List[Instruction]) -> (Dict[str, Any], CompilationMetrics):
```
hamidelmaazouz, line 76: "CompilationMetrics" sounds confusing when it's returned from a function that "execute"s instructions.
suggestion to rename `CompilationMetrics` to `QatMetric` as it's clearly not only about compilation metrics anymore now that you're adding new metrics.

src/QAT/qat/purr/compiler/execution.py
```diff
@@ -219,15 +224,16 @@ def _generate_repeat_batches(self, repeats):
         )
         return ([batch_limit] * list_expansion_ratio) + [repeats % batch_limit]
 
-    def execute(self, instructions):
+    def execute_with_metrics(self, instructions):
```
hamidelmaazouz, line 227: minor: hints for argument and return types please ^^

hamidelmaazouz (changes requested):
There's a pattern throughout the code changes that I think is not necessary:

```
context_manager = log_duration()
with context_manager:
    # Do stuff
some_metric = SomeMetric()
some_metric.record_metric(MetricsType, context_manager.duration)
```

I prefer that you put metric handling away where it belongs because `log_duration()` is capable of so much more:

```
def log_duration(message, metric_type=None, metric_collection=None, log_level: LoggerLevel = None)
```

This means you can tweak `log_duration` and/or develop `_LogContextManager` to be able to serve what you're trying to do with only one to two liner code changes at the call sites.

Issue #16, opened 2023-10-05
QASM get_phase and get_frequency result in errors


ayush-pancholy:
**Describe the bug**
When submitting QASM strings that contain `get_phase` or `get_frequency`, I get a `TypeError: 'PulseChannelView' object is not subscriptable`. 

**Describe what you expected to happen**
I expect to be able to use these QASM functions without error - it's nice to be able to query the device data without having to be sent it externally.

**Add supporting information**
As an example QASM snippet that fails, something as simple as querying the frequency of `q0_drive` and leaving it unchanged results in an error: `set_frequency(q0_drive, get_frequency(q0_drive));`

Full stack trace:
```Traceback (most recent call last):
  File "/home/ayushp/qat_playground/qat_qasm_example.py", line 87, in <module>
    builder2 = parser.parse(
  File "/home/ayushp/miniconda3/envs/qat/lib/python3.9/site-packages/qat/purr/integrations/qasm.py", line 2114, in parse
    self.visit(parsed)
  File "/home/ayushp/miniconda3/envs/qat/lib/python3.9/site-packages/lark/visitors.py", line 419, in visit
    return self._visit_tree(tree)
  File "/home/ayushp/miniconda3/envs/qat/lib/python3.9/site-packages/lark/visitors.py", line 427, in _visit_tree
    return f(tree)
  File "/home/ayushp/miniconda3/envs/qat/lib/python3.9/site-packages/lark/visitors.py", line 437, in __default__
    return self.visit_children(tree)
  File "/home/ayushp/miniconda3/envs/qat/lib/python3.9/site-packages/lark/visitors.py", line 430, in visit_children
    return [self._visit_tree(child) if isinstance(child, Tree) else child
  File "/home/ayushp/miniconda3/envs/qat/lib/python3.9/site-packages/lark/visitors.py", line 430, in <listcomp>
    return [self._visit_tree(child) if isinstance(child, Tree) else child
  File "/home/ayushp/miniconda3/envs/qat/lib/python3.9/site-packages/lark/visitors.py", line 427, in _visit_tree
    return f(tree)
  File "/home/ayushp/miniconda3/envs/qat/lib/python3.9/site-packages/qat/purr/integrations/qasm.py", line 1711, in quantum_gate_call
    self._call_gate(name, method_args)
  File "/home/ayushp/miniconda3/envs/qat/lib/python3.9/site-packages/qat/purr/integrations/qasm.py", line 1688, in _call_gate
[...]

PR #17, opened 2023-10-05
fix get phase and frequency

ayush-pancholy:
When sending QASM strings that contain `get_phase` and `get_frequency` instructions, the compiler produces an error. This is because `args` is not an array (since only one argument is provided) but rather a `PulseChannelView` and is therefore not subscriptable. Here's a simple fix that simply passes `args` instead of attempting to index into it first.

This fixes issue #16.


ayush-pancholy:

chemix-lunacy:
Looks good, can you please uncomment the `get_phase` and `get_frequency` calls in this test: https://github.com/oqc-community/qat/blob/main/src/tests/files/qasm/qasm3/lark_parsing_test.qasm#L69 to add a sanity check that your fix works. Additional tests may also be useful.

Looks like you fixed the user account issues on your second commit but the entire history needs to be like that. Just remove/re-add the commits with the new credentials, and with that you can also remove teh extrenious whitespace addition.

PR #19, opened 2023-10-18
Feature/ker/separate config

keriksson-rosenqvist:
This change allows us to distribute the config separately as well as provides improved backwards compatibility for serialised configs.


src/CONFIG/pyproject.toml
```diff
@@ -0,0 +1,24 @@
+[tool.poetry]
+name = "qat-config"
```
chemix-lunacy, line 2: Likely want it to be `qat-compiler-config` to make sure there's no confusion with other projects.

src/CONFIG/pyproject.toml
```diff
@@ -0,0 +1,24 @@
...
+    "Owen Arnold <oarnold@oxfordquantumcircuits.com>",
+]
+#readme = "README.md"
+packages = [{include = "qat_config"}]
+license = "BSD-3-Clause"
+
+[tool.poetry.dependencies]
+python = ">=3.8.1,<3.11"
+
+
```
chemix-lunacy, line 21: Space.

src/CONFIG/pyproject.toml
```diff
@@ -0,0 +1,24 @@
...
+authors = [
+    "Hamid El Maazouz <helmaazouz@oxfordquantumcircuits.com>",
+    "Harry Waring <hwaring@oxfordquantumcircuits.com>",
+    "Jamie Friel <jfriel@oxfordquantumcircuits.com>",
+    "John Dumbell <jdumbell@oxfordquantumcircuits.com>",
+    "Kajsa Eriksson Rosenqvist <keriksson.rosenqvist@oxfordquantumcircuits.com>",
+    "Norbert Deak <ndeak@oxfordquantumcircuits.com>",
+    "Owen Arnold <oarnold@oxfordquantumcircuits.com>",
+]
+#readme = "README.md"
```
chemix-lunacy, line 14: Remove commented stuff (for now).

chemix-lunacy (changes requested):
Should see the things we talked about changed in regards to proxy imports. Also I'd like to see some small changes here to do with structure:

`QAT/qat/purr` should hold the new poetry.toml/lock (it can probably just be in some random folder and use relative includes)
`QAT/qat/purr/external` should hold the files we need to expose externally. These should then be imported normally by the files they need proxying from.

We don't want QAT to import part of itself as a package as that just makes it way more awkward to develop against. We just want to _distribute_ parts of itself in isolation. With this everything stays almost the same except we just have a mini-module we send externally, and these files also have places in the project they live that just do a star import.

If there's problems with this and it dosen't work just after a simple move I'll have a quick look and try to quickly fix things up, as I think that setup is probably fine for the foreseeable future.

src/CONFIG/pyproject.toml
keriksson-rosenqvist: I was actually going to get this replaced with an actual README as that shows up on the PyPI page.

PR #20, opened 2023-11-03
Partial Package Builder Tool.

keriksson-rosenqvist:
Added a `PartialPackageBuilder` tool that allows easy building of partial packages from additional pyproject.toml files.

The additional files should have a different name, e.g. `xzy_pyproject.toml` and in addition to the normal poetry pyproject.toml fields they must include a `package_includes` list of package relative path strings to add to the partial package builds.

The tool is executed like `poetry run python tools/partial_package_builder.py -toml-file qat-compiler-minimal_pyproject.toml` and on successful completion, `tar.gz` and/or `whl` files of the partial package will be found in the `dist/` directory.

Long term, the intention is to make this into a proper Poetry plugin that we can use across our projects as needed.


chemix-lunacy:
Looks good, few improvements that should be done at some point:

1. The minimal .toml should inherit values off the main project. So it'd load the main project toml, then replace any values in it with the minimal one. Minimal .toml then just becomes the values you want changed. Saves duplication
2. The minimal README should live in a docs folder to clean things up, but it should also just make mention of what the project is for, then link directly to the main repo page, not just copy it wholesale.

Right now it just means whenever people make changes to the project version and documentation it needs to be made in two places. Or more likely people just forget to change the minimal one.

jfriel-oqc (approved):
Agree with John's suggestions on minimising duplication, but love this, great solution to our needs of config or qat builds! :ok_hand:

keriksson-rosenqvist:
Shortened the README, grabbing version from main poetry, and moving partial package definitions to a sub-dir for cleanliness.
Agree that long term we want to pull dependency versions etc. from main pyproject.toml file, however that becomes more complex to implement correctly and can be addressed in a future (plugin) version.

src/tests/test_config.py
```diff
@@ -21,6 +21,8 @@
 
 from tests.qasm_utils import TestFileType, get_test_file_path
 
+SUPPORTED_CONFIG_VERSIONS = ["v02", "v1"]
```
hamidelmaazouz, line 24: minor: is there a `"v01"` ? ^^

tools/partial_package_builder.py
```diff
@@ -0,0 +1,77 @@
...
+from tempfile import TemporaryDirectory
+
+import toml
+
+parser = argparse.ArgumentParser(
+    "Partial Package builder",
+    description="Easily package parts of the source code into a separate distribution."
+)
+parser.add_argument(
+    "-toml-file",
```
hamidelmaazouz, line 16: usually double dashes for long options

hamidelmaazouz (approved):

src/tests/test_config.py
keriksson-rosenqvist: Nope

PR #32, opened 2024-01-10
Allowing the hardware to work without a control hardware.

keriksson-rosenqvist:


chemix-lunacy (approved):
I'd like the additional dependency removed, but it's also harmless and will just be removed later anyway.

hamidelmaazouz (approved):

PR #33, opened 2024-01-11
Efficiency improvements

keriksson-rosenqvist:
Efficiency improvements:
- Efficient entanglement checker.
- Keep track of pulse channel duration as we go.
- Take the final instruction as our 'end' sample.
- Common add_channel method.
- Use sets for unique sets.
- Fix duplicate quantum targests.


chemix-lunacy (approved):
I'm approving on the assumption that these are all the changes in dev with no/minor modifications - since I'm pretty sure I recognize most of these.

If there are changes can you please point out which bits have changed. (Beyond small things like ordering/changing numpy to np, etc)

chemix-lunacy (approved):

PR #34, opened 2024-01-12
Update Copyright

hamidelmaazouz:


chemix-lunacy (approved):

keriksson-rosenqvist:
This should never have been updated. As per long discussions about copyright it's important to keep the first application date of the copyright as we otherwise will "loose protection" for things prior to the given date.

PR #35, opened 2024-01-12
Frequency Shift.

keriksson-rosenqvist:
Adding frequency shifting functionality.


src/QAT/qat/purr/compiler/execution.py
```diff
@@ -428,6 +428,19 @@ def create_duration_timeline(self, package: QatFile):
             if all(isinstance(pos_data.instruction, Delay) for pos_data in timeline):
                 del results[key]
 
+        if final_positions := [final_position[-1].end for final_position in list(results.values())]:
+            end_point = max(final_positions)
+
+            for qubit in self.model.qubits:
+                if freq_channel := qubit.pulse_channels.get("freq_shift", None):
```
chemix-lunacy, line 435: I forget what this is key'd on, but `ChannelType.freq_shift.name()` or similar will be better.

src/tests/test_freq_shift.py
```diff
@@ -0,0 +1,60 @@
+# SPDX-License-Identifier: BSD-3-Clause
```
chemix-lunacy, line 1: Either shift tests to another class or just called it `test_pulses` or something. You don't usually have a test file for a single feature unless it's amazingly critcal

src/tests/test_freq_shift.py
```diff
@@ -0,0 +1,60 @@
...
+        qubit1.add_pulse_channel(pulse_channel=freq_channel, channel_type=ChannelType.freq_shift)
+        hardware.add_pulse_channel(freq_channel,)
+
+    builder.pulse(
+        drive_channel_1, PulseShapeType.SQUARE, width=1e-6, amp=1, ignore_channel_scale=True
+    ).pulse(
+        drive_channel_2, PulseShapeType.SQUARE, width=1e-6, amp=1, ignore_channel_scale=True
+    )
+
+    qat_file = InstructionEmitter().emit(builder.instructions, builder.model)
```
chemix-lunacy, line 45: You want to create a fake engine and deal with all of this in the execute there. Look at some of the other tests for this.

src/QAT/qat/purr/compiler/devices.py
```diff
@@ -446,6 +447,24 @@ def __hash__(self):
...
+        self,
+        id_: str,
+        physical_channel: PhysicalChannel,
+        frequency=0.0,
+        bias=0.0 + 0.0j,
+        scale=1.0 + 0.0j,
+        amp=0.0,
+        active: bool = True,
+        fixed_if: bool = False,
+        **kwargs
```
chemix-lunacy, line 461: Better to just define the rest of the args here since you're a child (and already have for most of them)

chemix-lunacy (approved):
Please make the dictionary key something not just a string and make the tests use the usual runtime framework, but otherwise looks good. I was considering whether there's a better place for something like this, but duration timeline is probably appropriate.

chemix-lunacy (approved):

PR #39, opened 2024-01-17
Allow instructions to be returned as a metric

chemix-lunacy:
Knowing the optimized instructions is useful for many reasons but massively bloats the result, so make it an optional metric.

At this point the term metric may not be as appropriate as it used to be.

### DIFF
--- a/src/QAT/qat/purr/compiler/runtime.py
+++ b/src/QAT/qat/purr/compiler/runtime.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2024 Oxford Quantum Circuits Ltd
-
+import os
 from typing import List, TypeVar, Union
 
 from qat.purr.compiler.builders import InstructionBuilder, QuantumInstructionBuilder
@@ -119,6 +119,10 @@ class QuantumRuntime(MetricsMixin):
         self.record_metric(
             MetricsType.OptimizedInstructionCount, opt_inst_count := len(instructions)
         )
+
+        if self.are_metrics_enabled(MetricsType.OptimizedInstructions):
+            self.record_metric(MetricsType.OptimizedInstructions, os.linesep.join([str(inst) for inst in instructions]))
+
         log.info(f"Optimized instruction count: {opt_inst_count}")
 
         return self.engine.execute(instructions, results_format)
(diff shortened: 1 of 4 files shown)


src/tests/test_config.py
```diff
@@ -97,6 +101,29 @@ def find_subclasses(obj):
 
             assert vars(first_conf.optimizations) == vars(second_conf.optimizations)
 
+    def test_optimized_instructions_metric(self):
+        conf = CompilerConfig()
+        conf.metrics = conf.metrics | conf.metrics.OptimizedInstructions
```
hamidelmaazouz, line 106: would explicitly set it to `conf.metrics.OptimizedInstructions`, no chance for default metrics to interfere if they ever shift.

hamidelmaazouz: also default metrics are tested separately in the following test

src/QAT/qat/purr/compiler/runtime.py
```diff
@@ -119,6 +119,10 @@ def execute(
         self.record_metric(
             MetricsType.OptimizedInstructionCount, opt_inst_count := len(instructions)
         )
+
+        if self.are_metrics_enabled(MetricsType.OptimizedInstructions):
+            self.record_metric(MetricsType.OptimizedInstructions, os.linesep.join([str(inst) for inst in instructions]))
```
hamidelmaazouz, line 124: love the OS agnostic line sep ^^

hamidelmaazouz (approved):

src/tests/test_config.py
chemix-lunacy: Metrics will never interfere with each other, it doesn't test for them, and doing a flag append is the natural way you would enable the feature so testing that is meaningful. Not a bad suggestion though!

hamidelmaazouz: ah yes, my mind went straight to Python `or` ^^ the flag append operator makes sense :smiley:

jfriel-oqc (approved):

chemix-lunacy:
Converting to draft for now as there's a few things I want to do here before it's merged, so don't want an accidental merge.

chemix-lunacy:
Closing this for a few reasons: doubt I'll get to it anytime soon and the change is trivial if/when we ever want it.

PR #40, opened 2024-01-18
Add way to get most probable bitstring

chemix-lunacy:
Removed the binary count squash mechanism as it's not used, replaced with what binary was meant to be used for instead. Left in binary results form from a QPU because it can still have use in niche situations. One of those things that is greatly useful if you need it.


src/tests/test_qir.py
keriksson-rosenqvist, line 1: Why do the tests here need full runtime?

keriksson-rosenqvist (commented):
Looks good but why are the QIR tests skipped?

keriksson-rosenqvist (approved):
Looks good to me.

PR #43, opened 2024-01-18
[WIP] Misc improvements

hamidelmaazouz:
- Collect playback_results earlier to allow merging acq processing loops
- Record post-processing instructions in QatFile to avoid looping on instructions

### DIFF
--- a/src/QAT/qat/purr/compiler/emitter.py
+++ b/src/QAT/qat/purr/compiler/emitter.py
@@ -3,7 +3,7 @@
 
 from __future__ import annotations
 
-from typing import Any, List, Set
+from typing import Any, List, Set, Dict
 
 from qat.purr.compiler.instructions import (
     Acquire,
@@ -49,6 +49,7 @@ class QatFile:
     def __init__(self):
         self.timeline: List[TimelineSegment] = []
         self.meta_instructions = []
+        self._pp_instructions: Dict[str, List[PostProcessing]] = {}
 
     def add(self, *args, **kwargs):
         self.timeline.append(TimelineSegment(*args, **kwargs))
@@ -73,15 +74,13 @@ class QatFile:
         return [seg.instruction for seg in self.timeline]
 
     def get_pp_for_variable(self, target_var):
-        results = []
-        for instruction in self.instructions:
-            if (
-                isinstance(instruction, PostProcessing)
-                and instruction.acquire.output_variable == target_var
-            ):
-                results.append(instruction)
+        if target_var not in self._pp_instructions:
+            raise ValueError(f"Could not find PostProcessing instruction for variable {target_var}")
 
-        return results
+        return self._pp_instructions[target_var]
+
+    def add_pp_for_variable(self, pp, target_var):
+        self._pp_instructions.setdefault(target_var, []).append(pp)
 
     @property
     def sweeps(self):
@@ -113,6 +112,7 @@ class InstructionEmitter:
         for inst in instructions:
             if isinstance(inst, PostProcessing):
                 qatf.add(inst, [inst.acquire.channel.full_id()])
+                qatf.add_pp_for_variable(inst, inst.acquire.output_variable)
             elif isinstance(inst, QuantumInstruction):
                 qatf.add(inst)
             else:


src/QAT/qat/purr/backends/live.py
```diff
@@ -218,20 +222,9 @@ def _execute_on_hardware(self, sweep_iterator: SweepIterator, package: QatFile):
                         aq.mode == AcquireMode.INTEGRATOR
                     )
 
-            playback_results = self.model.control_hardware.start_playback(
```
chemix-lunacy, line 221: I'm going to need to see an in-depth breakdown on why this is safe considering you're pretty drastically changing the logic. I would also wager the speed increase is minute.

src/QAT/qat/purr/compiler/emitter.py
```diff
@@ -72,17 +73,6 @@ def add_meta(self, instruction):
     def instructions(self):
         return [seg.instruction for seg in self.timeline]
 
-    def get_pp_for_variable(self, target_var):
```
chemix-lunacy, line 75: Spirit of the change is fine, but I'd recommend you make the field private, put this method back in, it uses the dictionary and you also throw a meaningful error if the key dosen't exist.

It never failed before because if it found nothing it'd just return an empty dictionary.

chemix-lunacy (changes requested):
I'd like to see a more defensive post-processing change and a robust breakdown of why how you've changed the live engine won't break it as the logic has changed.

src/QAT/qat/purr/backends/live.py
```diff
@@ -218,20 +222,9 @@ def _execute_on_hardware(self, sweep_iterator: SweepIterator, package: QatFile):
...
-                repetitions=repetitions, repetition_time=repetition_time
-            )
-
-            for channel, aqs in aq_map.items():
-                if len(aqs) > 1:
-                    raise ValueError(
```
jfriel-oqc, line 227: why remove this check? This is currently a correct validation

jfriel-oqc (changes requested):
changes are fine up to the comments here

src/QAT/qat/purr/backends/live.py
hamidelmaazouz: it's still there, just from another loop ^^

src/QAT/qat/purr/backends/live.py
hamidelmaazouz: So the structure was like:

```
Loop 1: [for aqs in aq_map.values()]

Call + assignment: [playback_results = self.model.control_hardware.start_playback(repetitions=repetitions, repetition_time=repetition_time)]

Loop 2: for [aqs in aq_map.values()]
```

Two changes happened here:

1. Interchange `Loop 1` and `Call + assignment` so it becomes
```
Call + assignment: [playback_results = self.model.control_hardware.start_playback(repetitions=repetitions, repetition_time=repetition_time)]

Loop 1: [for aqs in aq_map.values()]

Loop 2: for [aqs in aq_map.values()]
```

This change is safe because the call `control_hardware.start_playback` invokes the control hardware while `Loop 1` just iterates through `aq_map` and writes to `SimpleAcquire.physical_channel`: both statements target completely different things from different layers and their side effects are isolated from one another.

2. Merge loops `Loop 1` and `Loop 2`
`Loop 1` and `Loop 2` are merged so it becomes:

Call + assignment: [playback_results = self.model.control_hardware.start_playback(repetitions=repetitions, repetition_time=repetition_time)]

Loop 3: [for aqs in aq_map.values()]
```

From a data dependence analysis, each iteration in `Loop 1` reads aq.mode and reads and writes to aq.physical_channel (+ trivial R/W operations), each iteration of `Loop 2` reads fields from `aq` (+ trivial R/W operations). Most importantly, the elements of `acq_map.values()` as well as the elements of `acqs` are independent of one another which makes the R/W effects of `Loop 1` + `Loop 2` linearly constructible and the proposed change is safe.

src/QAT/qat/purr/compiler/emitter.py
hamidelmaazouz: good point, `get_pp_for_variable` could be used by customers and this shouldn't break that, will push changes shortly.

chemix-lunacy (changes requested):
Once the loop changes are reverted I'll happily approve.

jfriel-oqc (approved):

PR #46, opened 2024-01-23
Bugfix/ker/batching fix

keriksson-rosenqvist-oqc:
Ensuring sweep iterators are always reverted and results can be combined for both `np.array` and `list` values.


chemix-lunacy (approved):
Assuming the other stuff is just additional fixes brought in?

jfriel-oqc (approved):

PR #41, opened 2024-01-18
Bugfix/ker/sequential qubit referencing

keriksson-rosenqvist:
Adjusting the way qubits references are handled in QASM and QIR integration to allow parsing and execution on hardware with jagged physical qubit indexing.

PR #61, opened 2024-02-07
Feature/ker/sequential qubit referencing

keriksson-rosenqvist:
Porting changes from main, #41


chemix-lunacy (approved):
Assuming this is just a direct port of the other PR? (No need for a re-review from my end if so)

PR #62, opened 2024-02-08
Bugfix/ker/minor fixes

keriksson-rosenqvist:
Collated minor fixes ported from main:
- Test raise TypeError when executing unsupported type.
- Enable skipped complex gate test.
- Ensure sweep iterators are always reverted.
- Extend test_load_hardware_definition to different qubit_counts, and temp fix reursion issue.
- Convert test_calibration.py to pytest format.
- Fix minor typo.
- Improved QuantumInstruction add_channel functionality.
- Removed duplicate numpy import.
- Allowing the hardware to work without a control hardware.
- Adding hardcoded delay to qubit drives and counting two clock cycles per instruction for the sequencer.
- Wh/switch hive fix.
- Fix frequency frame attribute access.

### DIFF
--- a/src/QAT/qat/purr/backends/live_devices.py
+++ b/src/QAT/qat/purr/backends/live_devices.py
@@ -74,10 +74,11 @@ class ControlHardwareChannel(PhysicalChannel):
     derived) object should contain hardware specific information.
     """
 
-    def __init__(self, id_, hardware_id, dcbiaschannel_pair, *args, **kwargs):
+    def __init__(self, id_, hardware_id, dcbiaschannel_pair, *args, switch_ch=None, **kwargs):
         super().__init__(id_, *args, **kwargs)
         self.hardware_id = hardware_id
         self.dcbiaschannel_pair: Dict[str, DCBiasChannel] = dcbiaschannel_pair
+        self.switch_ch: str = switch_ch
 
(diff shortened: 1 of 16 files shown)


src/QAT/qat/purr/backends/live_devices.py
```diff
@@ -74,10 +74,11 @@ class ControlHardwareChannel(PhysicalChannel):
     derived) object should contain hardware specific information.
     """
 
-    def __init__(self, id_, hardware_id, dcbiaschannel_pair, *args, **kwargs):
+    def __init__(self, id_, hardware_id, dcbiaschannel_pair, *args, switch_ch=None, **kwargs):
```
jfriel-oqc, line 77: this should be before *args

jfriel-oqc (approved):
approving up to that one argument order swtich

src/QAT/qat/purr/backends/live_devices.py
keriksson-rosenqvist: It's a `kwarg`, why would it be before `*args`?

jfriel-oqc: ah you're right, sorry! Misread with test wrapping

PR #66, opened 2024-02-23
Adding initial work for hybrid runtime

chemix-lunacy:


jfriel-oqc:
I think it's worth adding the remote execution engine here too, 99% of the time a hybrid runtime will be network connected and not directly plugged into a QPU (even if possible and we probably will for testing)

jfriel-oqc (changes requested):
remote engine requested for this to be usable and testable in prod systems

jfriel-oqc (approved):

poetry_scripts.py
```diff
@@ -10,6 +10,6 @@ def build_docs():
 
 def format_code():
     os.system("poetry run black .")
-    os.system("poetry run isort src/QAT")
-    os.system("poetry run isort src/tests")
+    # os.system("poetry run isort src/QAT")
```
jfriel-oqc, line 13: need to rebase to revert these bits now they're fixed

PR #67, opened 2024-02-27
Ported interrupt work

owen-oqc:
Allow wait interrupts to break batches


owen-oqc:

PR #77, opened 2024-03-12
Ported qiskit simulator strict placement checks

helmaazouz:


src/tests/test_qiskit_backend.py
```diff
@@ -16,6 +16,27 @@
 
 
 class TestQiskitBackend:
+
+    @pytest.mark.parametrize("strict_placement", [True])
```
keriksson-rosenqvist, line 20: Add `False`?

src/QAT/qat/purr/backends/echo.py
```diff
@@ -115,6 +116,14 @@ def generate_connectivity(con_type, qubit_count):
     return None
 
 
+def add_direction_couplings_to_hardware(model, connectivity):
```
keriksson-rosenqvist, line 119: This isn't echo specific. I know a few things around here aren't. Might be worth thinking about moving them somewhere else. :thinking:

keriksson-rosenqvist (approved):

src/QAT/qat/purr/backends/echo.py
helmaazouz: so not all setup functions add connectivity by default. For now the need rose for echo, once other needs come across, we can move to a common ground

PR #92, opened 2024-03-25
Bugfix/ker/measure instruction tests develop

keriksson-rosenqvist:
Enabling previously skipped tests for measurement instructions and fixing the issue that required it to be skipped.

### DIFF
--- a/src/tests/test_instructions.py
+++ b/src/tests/test_instructions.py
@@ -84,8 +84,18 @@ class TestInstruction:
 
 
 class TestInstructionExecution:
-    @pytest.mark.skip("Needs fixing post processing and down-conversion issues.")
-    def test_measure_instructions(hw):
+    @pytest.mark.parametrize(
+        "measure_instruction",
+        [
+            "measure_mean_z",
+            "measure_mean_signal",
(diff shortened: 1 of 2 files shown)


src/tests/test_instructions.py
```diff
@@ -84,8 +84,18 @@ def test_instruction_limit(self):
...
+    @pytest.mark.parametrize(
+        "measure_instruction",
+        [
+            "measure_mean_z",
+            "measure_mean_signal",
+            "measure_single_shot_z",
+            "measure_scope_mode",
+            "measure_single_shot_binned",
+            "measure_single_shot_signal",
+        ],
```
chemix-lunacy, line 96: For the future, it's better to use lambda's or method references than string -> getattr. Gets picked up by tools easier, refactors, fails with a meaningful message etc.

chemix-lunacy (approved):

jfriel-oqc (approved):

src/tests/test_instructions.py
keriksson-rosenqvist: @chemix-lunacy like so?

PR #95, opened 2024-03-27
Fix infinite recursion when disconnecting Instruments.

keriksson-rosenqvist:
Simple change that stops infinite looping between `Instrument.disconnect()` and `Instrument.driver = None`.

### DIFF
--- a/src/QAT/qat/purr/backends/live_devices.py
+++ b/src/QAT/qat/purr/backends/live_devices.py
@@ -36,7 +36,9 @@ class Instrument(Calibratable):
         if self.driver is not None:
             try:
                 self.driver.close()
-                self.driver = None
+                # The current version of the drivers required the link to be removed for
+                # proper operation. Adjust this when it is no longer needed.
+                self._driver = None
                 self.is_connected = False
             except BaseException as e:
                 log.warning(


src/QAT/qat/purr/backends/live_devices.py
```diff
@@ -36,7 +36,7 @@ def disconnect(self):
         if self.driver is not None:
             try:
                 self.driver.close()
-                self.driver = None
+                self._driver = None
```
hamidelmaazouz, line 39: I agree this breaks the call cycle `disconnect()` <-> `driver.setter()` but I don't think the setter should be called here at all.

Any idea why it's being called ?

keriksson-rosenqvist: Good point. :thinking:

keriksson-rosenqvist: Discussed with relevant parties and for the current driver versions this is required for proper operation.

jfriel-oqc: can we add a comment to this effect? I'll certainly forget this tomorrow

jfriel-oqc (approved):

src/QAT/qat/purr/backends/live_devices.py
hamidelmaazouz: on the other hand, the setter disconnects the driver also to potential reference owners. It isn't an issue now but might be in the future ...

hamidelmaazouz: Happy with the fix for the time being. For good measure, could we sweep at other usages of the setter and make sure to use the private field `_driver` directly from within `Instrument` ? I would like to keep the getter/setter is only for client structures

hamidelmaazouz (approved):
Happy with the fix for the time being. For good measure, could we look at other usages of the setter and make sure to use the private field _driver directly from within Instrument ? I would like to keep the getter/setter is only for client structures

PR #97, opened 2024-04-02
adding matrix testing for unit tests and code formatting

jfriel-oqc:


hamidelmaazouz (approved):
Thanks !

chemix-lunacy:
For Windows you'll need to add it to the path:

> To get started you need Poetry's bin directory (C:\Users\runneradmin\.local\bin) in your `PATH`
> environment variable.

Looks like Rasqals mac releases don't work with latest - I'll just do another release with latest on everything (I wanted to avoid it because latest is a moving goal).

Also I'd recommend running the various versions of Python on Linux, only doing one formatting check, and then running one version cross-OS. Pythons main differences across the OS' are paths and whatnot, and as long as your code works the version shouldn't matter.

That should cut down on amounts of runs.

PR #100, opened 2024-05-03
Bugfix/ker/instruction list fixes

keriksson-rosenqvist:
Ensure the generated instruction list is ordered consistently, needed for comparison in tests.

### DIFF
--- a/src/QAT/qat/purr/compiler/devices.py
+++ b/src/QAT/qat/purr/compiler/devices.py
@@ -494,6 +494,9 @@ class QubitCoupling(Calibratable):
         self.direction = tuple(direction)
         self.quality = 1 if quality < 1 else quality
 
+    def __repr__(self):
+        return f"{str(self.direction)} @{self.quality}"
+
 
 class PulseChannelView(PulseChannel):
     """
--- a/src/QAT/qat/purr/integrations/tket.py
+++ b/src/QAT/qat/purr/integrations/tket.py
@@ -397,8 +397,8 @@ def run_tket_optimizations(qasm_string, opts, hardware: QuantumHardwareModel) ->
         # trying to apply them to a limited subset of qubits provides no value.
         logical_qubit_map = {q.index: i for i, q in enumerate(hardware.qubits)}
         for c in couplings:
-            control_ind, targer_ind = c.direction
-            c.direction = (logical_qubit_map[control_ind], logical_qubit_map[targer_ind])
+            control_ind, target_ind = c.direction
+            c.direction = (logical_qubit_map[control_ind], logical_qubit_map[target_ind])
         if TketOptimizations.DefaultMappingPass not in opts:
             architecture = Architecture([val.direction for val in couplings])
             optimizations_failed = not optimize_circuit(circ, architecture, opts)
--- a/src/QAT/qat/purr/integrations/tket.py
+++ b/src/QAT/qat/purr/integrations/tket.py
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Oxford Quantum Circuits Ltd
+from copy import deepcopy
 from numbers import Number
 from typing import List
 
@@ -388,7 +389,7 @@ def run_tket_optimizations(qasm_string, opts, hardware: QuantumHardwareModel) ->
         )
         return qasm_string
 
-    couplings = hardware.qubit_direction_couplings.copy()
+    couplings = deepcopy(hardware.qubit_direction_couplings)
     optimizations_failed = False
     architecture = None
     if any(couplings):
--- a/src/QAT/qat/purr/integrations/tket.py
+++ b/src/QAT/qat/purr/integrations/tket.py
@@ -366,12 +366,16 @@ def run_tket_optimizations(qasm_string, opts, hardware: QuantumHardwareModel) ->
         )
         return qasm_string
 
-    couplings = hardware.qubit_direction_couplings
+    couplings = hardware.qubit_direction_couplings.copy()
     optimizations_failed = False
     architecture = None
     if any(couplings):
         # Without default remapping pass multi-qubit gates don't get moved around, so
         # trying to apply them to a limited subset of qubits provides no value.
+        logical_qubit_map = {q.index: i for i, q in enumerate(hardware.qubits)}
+        for c in couplings:
+            control_ind, targer_ind = c.direction
+            c.direction = (logical_qubit_map[control_ind], logical_qubit_map[targer_ind])
         if TketOptimizations.DefaultMappingPass not in opts:
             architecture = Architecture([val.direction for val in couplings])
             optimizations_failed = not optimize_circuit(circ, architecture, opts)
--- a/src/QAT/qat/purr/integrations/tket.py
+++ b/src/QAT/qat/purr/integrations/tket.py
@@ -396,9 +396,12 @@ def run_tket_optimizations(qasm_string, opts, hardware: QuantumHardwareModel) ->
         # Without default remapping pass multi-qubit gates don't get moved around, so
         # trying to apply them to a limited subset of qubits provides no value.
         logical_qubit_map = {q.index: i for i, q in enumerate(hardware.qubits)}
-        for c in couplings:
-            control_ind, target_ind = c.direction
-            c.direction = (logical_qubit_map[control_ind], logical_qubit_map[target_ind])
+        for coupling in couplings:
+            control_ind, target_ind = coupling.direction
+            coupling.direction = (
+                logical_qubit_map[control_ind],
+                logical_qubit_map[target_ind],
+            )
         if TketOptimizations.DefaultMappingPass not in opts:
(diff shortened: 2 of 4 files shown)


src/QAT/qat/purr/integrations/tket.py
```diff
@@ -366,12 +366,16 @@ def run_tket_optimizations(qasm_string, opts, hardware: QuantumHardwareModel) ->
         )
         return qasm_string
 
-    couplings = hardware.qubit_direction_couplings
+    couplings = hardware.qubit_direction_couplings.copy()
```
hamidelmaazouz, line 369: why copy?

src/QAT/qat/purr/integrations/tket.py
```diff
@@ -366,12 +366,16 @@ def run_tket_optimizations(qasm_string, opts, hardware: QuantumHardwareModel) ->
...
     optimizations_failed = False
     architecture = None
     if any(couplings):
         # Without default remapping pass multi-qubit gates don't get moved around, so
         # trying to apply them to a limited subset of qubits provides no value.
+        logical_qubit_map = {q.index: i for i, q in enumerate(hardware.qubits)}
+        for c in couplings:
+            control_ind, targer_ind = c.direction
```
hamidelmaazouz, line 377: you mean target ?

hamidelmaazouz: also, maybe they're just names but I advise against assigning control and target semantics to the tuple as you don't know which is which

src/QAT/qat/purr/compiler/devices.py
```diff
@@ -494,6 +494,9 @@ def __init__(self, direction, quality=1):
         self.direction = tuple(direction)
         self.quality = 1 if quality < 1 else quality
 
+    def __repr__(self):
```
hamidelmaazouz, line 497: I don't understand its use :thinking:

src/QAT/qat/purr/integrations/tket.py
keriksson-rosenqvist: Swapped to `deepcopy`. It's because we do not want to actually change the couplings on the `hardware model`.

src/QAT/qat/purr/integrations/tket.py
keriksson-rosenqvist: You do know which i which, in the sense that the coupling quality is based on the specific direction. Will fix typo.

src/QAT/qat/purr/integrations/tket.py
```diff
@@ -366,37 +389,33 @@ def run_tket_optimizations(qasm_string, opts, hardware: QuantumHardwareModel) ->
         )
         return qasm_string
 
-    couplings = hardware.qubit_direction_couplings
+    couplings = deepcopy(hardware.qubit_direction_couplings)
```
jfriel-oqc, line 392: why does this need deep copying?

src/QAT/qat/purr/integrations/tket.py
```diff
@@ -366,37 +389,33 @@ def run_tket_optimizations(qasm_string, opts, hardware: QuantumHardwareModel) ->
...
     optimizations_failed = False
     architecture = None
     if any(couplings):
         # Without default remapping pass multi-qubit gates don't get moved around, so
         # trying to apply them to a limited subset of qubits provides no value.
+        logical_qubit_map = {q.index: i for i, q in enumerate(hardware.qubits)}
+        for c in couplings:
```
jfriel-oqc, line 399: c is a bad variable name

src/QAT/qat/purr/integrations/tket.py
```diff
@@ -366,37 +389,33 @@ def run_tket_optimizations(qasm_string, opts, hardware: QuantumHardwareModel) ->
...
         return qasm_string
 
-    couplings = hardware.qubit_direction_couplings
+    couplings = deepcopy(hardware.qubit_direction_couplings)
     optimizations_failed = False
     architecture = None
     if any(couplings):
         # Without default remapping pass multi-qubit gates don't get moved around, so
         # trying to apply them to a limited subset of qubits provides no value.
+        logical_qubit_map = {q.index: i for i, q in enumerate(hardware.qubits)}
```
jfriel-oqc, line 398: this is a property on the HW already. We don't want to be redefining this property any more than once, that'll let the demons in

src/QAT/qat/purr/integrations/tket.py
```diff
@@ -346,6 +347,28 @@ def optimize_circuit(circ, architecture, opts):
         return False
 
 
+def get_coupling_subgraphs(couplings):
```
jfriel-oqc, line 350: this needs its own set of tests

jfriel-oqc (changes requested):

src/QAT/qat/purr/integrations/tket.py
keriksson-rosenqvist: It's because we do not want to actually change the couplings on the hardware model.

src/QAT/qat/purr/integrations/tket.py
keriksson-rosenqvist: No it's not.

src/QAT/qat/purr/integrations/tket.py
keriksson-rosenqvist: Added

src/QAT/qat/purr/integrations/tket.py
```diff
@@ -380,27 +403,19 @@ def run_tket_optimizations(qasm_string, opts, hardware: QuantumHardwareModel) ->
...
-                    continue
-
-                # Attempt optimization with good quality couplings, then degrade every
-                # failed try until we use everything.
-                last_couplings = len(filtered_couplings)
-                architecture = Architecture(filtered_couplings)
-                optimizations_failed = not optimize_circuit(circ, architecture, opts)
+            coupling_qualities = list({val.quality for val in couplings})
+            coupling_qualities.sort(reverse=True)
+            for quality_level in coupling_qualities:
```
hamidelmaazouz, line 411: scenario when qualities are pairwise different: First iteration will consider **only one** coupling and subsequent iterations will include one more coupling at a time until optimisation succeeds. While this is correct, the filtering in this setup is at its finest granularity and I think `optimizations_failed` will remain `True` until you include enough couplings to actually make sth happen on the circuit.

hamidelmaazouz: > scenario when qualities are pairwise different: First iteration will consider **only one** coupling and subsequent iterations will include one more coupling at a time until optimisation succeeds. While this is correct, the filtering in this setup is at its finest granularity and I think `optimizations_failed` will remain `True` until you include enough couplings to actually make sth happen on the circuit.

Previous code did the batching in 3 levels with which tket opt had a higher chance to "converge" quickly

src/QAT/qat/purr/integrations/tket.py
hamidelmaazouz: > You do know which is which, in the sense that the coupling quality is based on the specific direction

Oh I mean logically speaking, control and target semantics usually come from the circuit. I guess it's also fine with hw arch ...

src/QAT/qat/purr/compiler/devices.py
hamidelmaazouz: oh absolutely, I was under the impression it had a role in the actual fix itself ^^

src/QAT/qat/purr/integrations/tket.py
keriksson-rosenqvist: I don't understand what the question/point is.

src/QAT/qat/purr/integrations/tket.py
keriksson-rosenqvist: @hamidelmaazouz A quick search says there isn't one, at least not one that wouldn't require adding additional dependencies for just this feature. If we find one it can be replaced in a future PR. :slightly_smiling_face:

src/tests/test_tket.py
```diff
@@ -1,13 +1,38 @@
...
 from tests.qasm_utils import get_qasm2
 
 
+@pytest.mark.parametrize(
+    "couplings,subgraphs",
+    [
+        pytest.param(
+            [(0, 1), (1, 2), (2, 3), (4, 3)],
+            [[(0, 1), (1, 2), (2, 3), (4, 3)]],
+            id="Continuous",
```
jfriel-oqc, line 23: id is a nice add to parametrize tests, never actually used it before tbh!

src/QAT/qat/purr/integrations/tket.py
jfriel-oqc: ah yeah, fair point, that's an internal qat thing currently. Can we get a TODO here just to remember to consolidate once we go back over all this stiff

jfriel-oqc (approved):

src/QAT/qat/purr/integrations/tket.py
hamidelmaazouz: yeah basically increasing the batch size would help a lot

PR #103, opened 2024-05-10
Ported hardware connectivity concept from develop.

keriksson-rosenqvist:
Porting the connectivity concept and associated `apply_setup_to_hardware` improvements from develop. These upgrades will allow for more extensive testing to be done with multiple hardware setups in upcoming work.

**N.B.** This is still backwards compatible as the all-to-all cross resonance channels and default connectivity settings are set to generate the same hardware as before. This is different from the develop version.

### DIFF
--- a/src/QAT/qat/purr/backends/echo.py
+++ b/src/QAT/qat/purr/backends/echo.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Oxford Quantum Circuits Ltd
-from typing import Dict
+from enum import Enum, auto
+from typing import Dict, List, Optional, Tuple, Union
 
 import numpy as np
 from qat.purr.backends.utilities import get_axis_map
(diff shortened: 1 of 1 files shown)


src/QAT/qat/purr/backends/echo.py
```diff
@@ -52,32 +55,98 @@ def apply_setup_to_hardware(hw, qubit_count: int = 4):
...
+                        frequency=5.5e9,
+                        scale=50,
+                    )
+                    qubit.create_pulse_channel(
+                        auxiliary_devices=[other_qubit],
+                        channel_type=ChannelType.cross_resonance_cancellation,
+                        frequency=5.5e9,
+                        scale=0.0,
+                    )
+                qubit.add_coupled_qubit(qubit_devices[(i + 1) % qubit_count])
```
keriksson-rosenqvist, line 75: This is different from develop, in order to retain backwards compatibility.

src/QAT/qat/purr/backends/echo.py
```diff
@@ -52,32 +55,98 @@ def apply_setup_to_hardware(hw, qubit_count: int = 4):
...
+    return model
+
+
+class Connectivity(Enum):
+    Ring = auto()
+
+
+def get_default_echo_hardware(
+    qubit_count=4,
+    connectivity: Optional[Union[Connectivity, List[Tuple[int, int]]]] = None,
```
keriksson-rosenqvist, line 139: Default `connectivity` set to `None` in order to retain backwards compatibility.

jfriel-oqc (approved):

PR #104, opened 2024-05-13
Add in batching for optimizations

chemix-lunacy:
When you have larger machines and higher quality level variance runtime far outweighs any minor improvements made by trying to iterate through every single quality level. This also includes the fact that most people will be running larger circuits that use the majority of the QPU so we cut out large chunks of iterations which can never work due to lack of number of qubits.


src/QAT/qat/purr/integrations/tket.py
```diff
@@ -407,8 +407,11 @@ def run_tket_optimizations(qasm_string, opts, hardware: QuantumHardwareModel) ->
             optimizations_failed = not optimize_circuit(circ, architecture, opts)
         else:
             coupling_qualities = list({val.quality for val in couplings})
-            coupling_qualities.sort(reverse=True)
-            for quality_level in coupling_qualities:
+            min_quality = min(coupling_qualities)
```
keriksson-rosenqvist, line 410: For some live hardware this does not select the best qubits successfully. We need to hold off on merging this until that is resolved.

keriksson-rosenqvist (changes requested):

src/QAT/qat/purr/integrations/tket.py
chemix-lunacy: There's also a trade-off between 'perfect' and 'usable' if one takes 50 years to process. There's next to no difference between the two algorithms except this one just batches numbers together.

I'd like to see some examples of where it dosen't pick good-enough qubits and the quality settings have been set correctly in the model.

chemix-lunacy:
Closing for now, if we need to apply a patch later the change is small enough - and we know what it is - to deal with it then.

PR #111, opened 2024-06-13
Feature/new serialiser backend

jfriel-oqc:
Adding the new serialiser backend with some tests to verify that all is good with maintaining integrity of floats etc as that's a common trip up with some serialiser backends 

Added a little speed test script, too, nothing sophisticated, but something we can lean on for new serialisation, worth keeping in the tools folder imo

serialiser_test


src/tests/test_serializer.py
```diff
@@ -222,3 +230,72 @@ def test_deserialize_complex(self):
...
+        string = builder.serialize(backend=backend)
+
+        new_builder = InstructionBuilder.deserialize(string, backend=backend)
+
+        for instruction in new_builder.instructions:
+            if isinstance(instruction, PhaseShift):
+                phase = instruction.phase
+                if isclose(phase, self.floating_point_num):
+                    # There are phase instructions that aren't just the pure rotation, easiest way to find them
+                    assert instruction.phase == self.floating_point_num
```
owen-oqc, line 287: This is a rare exception to using a == operator. However it might be more interesting to convert to a bit sequence and compare that. At least you can then go off and debug what the serialisation library has done. For example https://pypi.org/project/bitstring/

owen-oqc (commented):
Kudos for getting data together not just comparing a single point before and after.

src/QAT/qat/purr/compiler/builders.py
```diff
@@ -72,18 +77,25 @@ def instructions(self):
         return list(self._instructions)
 
     @staticmethod
-    def deserialize(blob) -> "InstructionBuilder":
+    def deserialize(
+        blob, backend: SerialiserBackend = SerialiserBackend.json
```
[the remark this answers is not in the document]
jfriel-oqc: this is the same as old, just being explicit to use the default backend rather than leaving it as it can now be set

src/tests/test_serializer.py
```diff
@@ -222,3 +230,72 @@ def test_deserialize_complex(self):
...
+
+
+@pytest.mark.parametrize("backend", [SerialiserBackend.json, SerialiserBackend.ujson])
+class TestBuilderSerialiser:
+    """
+    Note here, that typically doing a direct float comparison is not meaningful in a test.
+    Here however, we need to ensure that the integrity of the floats is exactly preserved across serialisation.
+    """
+
+    floating_point_num = math.sqrt(math.pi) / 2.14324354
```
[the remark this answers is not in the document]
jfriel-oqc: good point, this is a bit lazy. I'll expand.

The key point is the risk that floats are dealt with slightly differently by different backends, especially if the language the serialiser is written is different. It's essential we precisely respect floats

src/tests/test_serializer.py
jfriel-oqc: I don't quite understand the value of this. We're only seeing the effect of the final result in python so would see that a bit string is different or a float is different. This is effectively a backstop test of a third party library that I'd usually not both with, but just need this for integrity checking

keriksson-rosenqvist:
Any particular reason this work is done towards develop and not main?

jfriel-oqc:
> Any particular reason this work is done towards develop and not main?

because I didn't think, I just started a new branch off dev. Will do our branch switching asap

src/tests/test_serializer.py
```diff
@@ -222,3 +230,90 @@ def test_deserialize_complex(self):
...
+        """
+        builder = builder(floating_point_num)
+        string = builder.serialize(backend=backend)
+
+        new_builder = InstructionBuilder.deserialize(string, backend=backend)
+
+        assert (
+            new_builder.model.get_qubit(0).get_drive_channel().frequency
+            == floating_point_num
+        )
```
hamidelmaazouz, line 285: minor: add tests for the U gate parameters

src/tests/test_serializer.py
```diff
@@ -222,3 +230,90 @@ def test_deserialize_complex(self):
...
+        Testing that different serialisation backends respect floating point numbers in hardware
+
+        """
+        builder = builder(floating_point_num)
+        string = builder.serialize(backend=backend)
+
+        new_builder = InstructionBuilder.deserialize(string, backend=backend)
+
+        for instruction in new_builder.instructions:
+            if isinstance(instruction, PhaseShift):
```
hamidelmaazouz, line 300: minor: looks like you're targetting only phase_shifts in this test, could you rename the test accordingly ?

src/tests/test_serializer.py
```diff
@@ -222,3 +230,90 @@ def test_deserialize_complex(self):
...
+                    assert instruction.phase == floating_point_num
+
+    def test_object_reconstruction(self, backend, floating_point_num, builder):
+        builder = builder(floating_point_num)
+        string = builder.serialize(backend=backend)
+        new_builder = InstructionBuilder.deserialize(string, backend=backend)
+        model = new_builder.model
+
+        qubit_route_1 = model.get_qubit(0)
+        print(model.physical_channels)
```
hamidelmaazouz, line 313: rogue print

tools/serialiser_speed.py
```diff
@@ -0,0 +1,58 @@
...
+
+def serialisation_time(builder, backend: SerialiserBackend):
+    """
+    Time isn't the best way to do this, but it's simple for the purposes of this test.
+    """
+    start = time.time()
+    string = builder.serialize(backend=backend)
+    reconstitued = InstructionBuilder.deserialize(string, backend=backend)
+    end = time.time()
+    return end - start
```
hamidelmaazouz, line 38: Not super critical for CPython implementations, but in general you want to run multiple times and avg the last few runs

hamidelmaazouz (commented):
Could you add cross-backend tests ? (Serialise with A and deserialise with B). Interest is in correctness and commutativity of different paths (if possible of course)

PR #114, opened 2024-06-19
Bug fix - Compare method calls instead of methods themselves

helmaazouz:


hamidelmaazouz (approved):

keriksson-rosenqvist (changes requested):
Looks good but can we add tests for this working properly on jagged hardware models as well?

keriksson-rosenqvist (approved):

jfriel-oqc (approved):

PR #115, opened 2024-06-20
Wh/changes for weighted readback

keriksson-rosenqvist:
A few changes to get weighted readback working

- Changes to acquire instruction to allow CustomPulses to be used
- New measure acquire dictionary entries to store weights for integrated readback
- Changes to measure to try and use weighted readback if wanted

### DIFF
--- a/src/QAT/qat/purr/compiler/builders.py
+++ b/src/QAT/qat/purr/compiler/builders.py
@@ -572,6 +572,7 @@ class QuantumInstructionBuilder(InstructionBuilder):
 
         measure_channel = qubit.get_measure_channel()
         acquire_channel = qubit.get_acquire_channel()
+        weights = qubit.measure_acquire.get('weights', None) if qubit.measure_acquire.get('use_weights', False) else None
         acquire_instruction = Acquire(
             acquire_channel,
             qubit.pulse_measure["width"]
@@ -581,6 +582,7 @@ class QuantumInstructionBuilder(InstructionBuilder):
             output_variable,
             self.existing_names,
             qubit.measure_acquire["delay"],
+            weights,
         )
 
         # If we detect a full measure block before us, merge it together if we're
--- a/src/QAT/qat/purr/compiler/instructions.py
+++ b/src/QAT/qat/purr/compiler/instructions.py
@@ -345,22 +345,16 @@ class Acquire(QuantumComponent, QuantumInstruction):
         self.mode: AcquireMode = mode or AcquireMode.RAW
         self.delay = delay
         self.output_variable = output_variable or self.generate_name(existing_names)
+        self.filter: Union[Pulse, CustomPulse] = self._check_filter(filter)
 
+    def _check_filter(self, filter):
         if filter is not None:
-            if isinstance(filter, (list, np.ndarray)):
-                filter = CustomPulse(channel, filter)
-            elif not isinstance(filter, (Pulse, CustomPulse)):
-                raise ValueError(
-                    "Filter on an acquire has to be a Pulse. Instead it's a "
-                    f"{type(filter)}"
-                )
-
             if not np.isclose(filter.duration, self.time):
                 raise ValueError(
                     f"Filter duration '{filter.duration}' must be equal to Acquire "
                     f"duration '{self.time}'."
                 )
-        self.filter: Union[Pulse, CustomPulse] = filter
+        return filter
 
     def generate_name(self, existing_names=None):
         return build_generated_name(existing_names, f"{self.channel.id}")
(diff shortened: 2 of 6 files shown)


hamidelmaazouz (approved):
Do we want a version bump in this PR itself ?

src/QAT/qat/purr/compiler/instructions.py
```diff
@@ -337,28 +337,27 @@ def __init__(
         output_variable=None,
         existing_names: Set[str] = None,
         delay=None,
-        filter: Pulse = None,
+        filter: Union[Pulse, CustomPulse] = None
```
hamidelmaazouz, line 340: same as `Waveform`

src/QAT/qat/purr/compiler/instructions.py
```diff
@@ -337,28 +337,27 @@ def __init__(
...
-                    "Filter on an acquire has to be a Pulse. Instead it's a "
-                    f"{type(filter)}"
-                )
-
-            if filter.duration != self.time:
-                raise ValueError(
-                    f"Filter duration '{filter.duration}' must be equal to Acquire "
-                    f"duration '{self.time}'."
-                )
+        self.filter: Union[Pulse, CustomPulse] = self._check_filter(filter)
```
hamidelmaazouz, line 348: same here ^^

src/QAT/qat/purr/compiler/instructions.py
```diff
@@ -337,28 +337,27 @@ def __init__(
...
+
+    def _check_filter(self, filter):
+        if filter:
+            for target in self.quantum_targets:
+                if isinstance(target, PulseChannel):
+                    dt = target.physical_channel.sample_time
+                    if not np.isclose(filter.duration, dt * (self.time // dt), atol=1e-12):
```
hamidelmaazouz, line 355: why `dt * (self.time // dt)` ?
use the systematic function `calculate_duration`
(@norbert-deak another reason why this function need not be an instance method)

hamidelmaazouz (changes requested):

src/QAT/qat/purr/compiler/instructions.py
norbert-deak: @hamidelmaazouz not sure I understand your comment here, might be missing some context

src/QAT/qat/purr/compiler/instructions.py
willh727-OQC: Assuming it was Pulse so AbstractWaveform couldn't be used? I actually have no ideal what AbstractWaveform is used for, but happy to change this to Waveform

src/QAT/qat/purr/compiler/builders.py
```diff
@@ -572,6 +572,7 @@ def measure(
 
         measure_channel = qubit.get_measure_channel()
         acquire_channel = qubit.get_acquire_channel()
+        weights = qubit.measure_acquire.get('weights', None) if qubit.measure_acquire.get('use_weights', False) else None
```
hamidelmaazouz, line 575: not a big deal but this smells like duplicate `None` semantics. Consider the following:

- `None` means do not use weights
- anything else means use weights, and "the thing" as weights

to avoid an additional `use_weights` entry ?

src/QAT/qat/purr/compiler/instructions.py
hamidelmaazouz: `Waveform` is the right type information that satisfies needs expressed here as its descendents are exactly `Pulse` and `CustomPulse`

hamidelmaazouz: I agree `AbstractWaveform` is a little confusing ... but that's another issue ^^

src/QAT/qat/purr/compiler/instructions.py
hamidelmaazouz: logic here basically computes duration based on the channel's sample time, if I remember correctly, it's the same logic that's been specified at `QuantumExecutionEngine.calculate_duration()`

hamidelmaazouz (changes requested):

src/QAT/qat/purr/compiler/builders.py
willh727-OQC: Not sure if I follow? Basically there's two ways to turn weights off, if it's set to None or if use_weights is set to False. Weights mirrors what loaded into memory for demolition and 'use_weights' allows you to turn it off and on without loosing the array. I find flexibility when doing calibrations is helpful for people.
Is this what you ment?

hamidelmaazouz: > without loosing the array

that's the keyword I needed 😄 just wanted to point it out, happy to hear this choice was made deliberately ^^

src/QAT/qat/purr/compiler/instructions.py
willh727-OQC: Ok I've made it a bit clearer, I was trying to recreate the round error but with np close properly set, its not needed. Also calculate duration isn't needed as duration is calculated inside the CustomPulse

Issue #117, opened 2024-06-24
Error executing in example/plot_pulse_buffers.py 


dnolivieri:
**Describe the bug**

Given a clean install with conda,  using pyhthon 3.9,  and install the qat via pip, 
the following error occurs when executing from  command line:  python plot_pulse_buffers.py
batch_results = self._execute_on_hardware(switerator, qat_file, interrupt)
TypeError: _execute_on_hardware() takes 3 positional arguments but 4 were given

**Describe what you expected to happen**
To override the bug,  I forced an extra parameter with this function:  
def _execute_on_hardware(self, sweep_iterator: SweepIterator, package: QatFile, interrupt=None):

This produces 
  File "/home/david/anaconda3/envs/quant39/lib/python3.9/site-packages/qat/purr/compiler/execution.py", line 448, in create_duration_timeline
    for instruction in instructions:
TypeError: 'QatFile' object is not iterable

**Add supporting information**

1. OS/version:  Ubuntu: 22.04;  conda 
2:  configuration of qat out of the box on the develop branch; no other changes.

hamidelmaazouz:
Thank you @dnolivieri,

We'll have a look and circle back.

hamidelmaazouz:
@dnolivieri,

Regarding this bit, I only see three arguments given to `_execute_on_hardware`.
```
batch_results = self._execute_on_hardware(switerator, qat_file, interrupt)
TypeError: _execute_on_hardware() takes 3 positional arguments but 4 were given
```

It looks to me that `_execute_on_hardware` wasn't being called properly, could you share contents of your script `plot_pulse_buffers.py` ?

Thanks,

dnolivieri:
Hi Hamid.   
Sure, but I didn't do really anything more than what is in the script on the develop branch, with some prints.
In order to be able to call the _execute_on_hardware,  I added a parameter interrupt=None, so that it would at least get into the function:  


 --- 
    def _execute_on_hardware(self, sweep_iterator: SweepIterator, package: QatFile, interrupt=None):
        position_map = self.target.create_duration_timeline(package)
        pulse_buffers = self.target.build_pulse_channel_buffers(
            position_map, do_upconvert=self.upconvert
        )
        buffers = self.target.build_physical_channel_buffers(pulse_buffers)

    --- >  rest of the code the same. 
if __name__ == "__main__":
    qasm = """
    OPENQASM 3;
    include "qelib1.inc";
    defcalgrammar "openpulse";
    qreg q[2];
[...]

PR #119, opened 2024-06-25
Fix plot pulse buffer example script

hamidelmaazouz:
A fix for https://github.com/oqc-community/qat/issues/117. The method `PhysicalBufferPlotEngine._execute_on_hardware()` should call `QuantumExecutionEngine.create_duration_timeline()` with the instructions list and not qat_file.


jfriel-oqc (approved):

keriksson-rosenqvist (commented):
What's the reason for this change? It seems to be breaking things that were previously working. :confused:

PR #122, opened 2024-06-26
Allow configuration of LiveHardwareModel on initistion.

keriksson-rosenqvist:
Allow `LiveHardwareModel` to be initiated with all the input arguments supported by `QuantumHardwareModel`.

### DIFF
--- a/src/QAT/qat/purr/backends/live.py
+++ b/src/QAT/qat/purr/backends/live.py
@@ -1,6 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Oxford Quantum Circuits Ltd
-from typing import Dict, List
+from typing import Dict, List, Optional
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


hamidelmaazouz:
Question: why do we need the extra arguments ? Are they required for some use case ?

hamidelmaazouz (commented):
Question: why do we need the extra arguments ? Are they required for some use case ?

norbert-deak (approved):

keriksson-rosenqvist:
> Question: why do we need the extra arguments ? Are they required for some use case ?

@hamidelmaazouz What are you referring to? This is allowing users of `LiveHardwareModel` to specify the same input arguments as are supported by the parent class `QuantumHardwareModel`. If this is not added then the `LiveHardwareModel` is "forced" to use the default `QuantumHardwareModel` values for those input arguments.

hamidelmaazouz (approved):

PR #123, opened 2024-06-27
Pre-commit and CI additions

keriksson-rosenqvist:
Tooling for easier format consistency.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -13,6 +13,29 @@ permissions:
 run-name: Build from ${{ github.ref }}
 
 jobs:
+  formatting-tests:
+    name: Consistency and formatting
+    runs-on: [ubuntu-latest]
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v3
+        with:
+          python-version: "3.9"
+
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+
+      - name: Formatting tests
+        run: |
+          poetry install --with dev
+          poetry run black --check .
+          poetry run isort --check .
+          poetry run autoflake --check .
+
+
   unit-tests:
     name: Unit Tests
     runs-on: [ubuntu-latest]
--- a/poetry_scripts.py
+++ b/poetry_scripts.py
@@ -1,0 +1,7 @@
+import os
+
+
+def format_code():
+    os.system("poetry run black .")
+    os.system("poetry run isort .")
+    os.system("poetry run autoflake .")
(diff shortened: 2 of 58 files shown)


pyproject.toml
```diff
@@ -78,11 +79,21 @@ build-backend = "poetry.core.masonry.api"
...
+in-place = true
+recursive = true
 
 [tool.isort]
-py_version = 38
-profile = "wemake"
+py_version = 39
+profile = "black"
 line_length = 92
+known_first_party = ["qat"]
```
hamidelmaazouz, line 89: we could also add these options as `args : []` in the pre-commit config file

pyproject.toml
```diff
@@ -78,11 +79,21 @@ build-backend = "poetry.core.masonry.api"
 remove-all-unused-imports = true
 ignore-init-module-imports = true
 ignore-pass-after-docstring = true
+in-place = true
+recursive = true
```
hamidelmaazouz, line 83: an example for pre-commit hook args for autoflake `args: [ --remove-all-unused-imports, --ignore-init-module-imports, --ignore-pass-after-docstring, --in-place ]`

.github/workflows/build.yaml
```diff
@@ -13,6 +13,29 @@ permissions:
...
+
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+
+      - name: Formatting tests
+        run: |
+          poetry install --with dev
+          poetry run black --check .
+          poetry run isort --check .
+          poetry run autoflake --check .
```
hamidelmaazouz, line 36: tests here also need to use the same flags used in pre-commit config file and pyproject (especially line-length)

poetry_scripts.py
```diff
@@ -0,0 +1,7 @@
+import os
+
+
+def format_code():
+    os.system("poetry run black .")
+    os.system("poetry run isort .")
+    os.system("poetry run autoflake .")
```
hamidelmaazouz, line 7: same idea here, we could make sure the same flags are consistent across different locations:
- this script
- pre-commit config
- poetry tooling config
- CI tests

hamidelmaazouz (approved):
Lovely addition Kajsa ! We could also apply these to test source code as well

pyproject.toml
```diff
@@ -78,11 +79,21 @@ build-backend = "poetry.core.masonry.api"
...
-py_version = 38
-profile = "wemake"
+py_version = 39
+profile = "black"
 line_length = 92
+known_first_party = ["qat"]
+
+[tool.black]
+line-length = 92
+target-version = ["py38", "py39", "py310"]
```
hamidelmaazouz, line 93: super minor: I believe black can infer which python version from project settings, if not it does it per file basis

hamidelmaazouz (approved):

poetry_scripts.py
keriksson-rosenqvist: I have verified that all these different executions make use of the settings defined in `pyproject.toml`.

hamidelmaazouz: thanks

PR #125, opened 2024-07-01
Imports from collections

keriksson-rosenqvist:
Adjust imports that have been deprecated in py3.10 while retaining compatibility with py3.8.

### DIFF
--- a/src/QAT/qat/purr/compiler/builders.py
+++ b/src/QAT/qat/purr/compiler/builders.py
@@ -2,7 +2,7 @@
 # Copyright (c) 2023 Oxford Quantum Circuits Ltd
 import itertools
 import math
-from collections import defaultdict
+from collections.abc import defaultdict
 from enum import Enum, auto
 from typing import List, Set, Union
 
--- a/src/QAT/qat/purr/compiler/builders.py
+++ b/src/QAT/qat/purr/compiler/builders.py
@@ -2,7 +2,12 @@
 # Copyright (c) 2023 Oxford Quantum Circuits Ltd
 import itertools
 import math
-from collections.abc import defaultdict
+import sys
+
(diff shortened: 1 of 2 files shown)


minsu-seo-oqc:

minsu-seo-oqc (approved):

hamidelmaazouz (approved):

src/QAT/qat/purr/compiler/builders.py
```diff
@@ -2,7 +2,12 @@
 # Copyright (c) 2023 Oxford Quantum Circuits Ltd
 import itertools
 import math
-from collections import defaultdict
+import sys
+
+if sys.version_info < (3, 10):
```
jfriel-oqc, line 7: why do we need this?

```
kristianb@jamie-xps-15-9520:~/Documents/code/scrapyard/pyt39_tes$ poetry run python
Python 3.8.10 (default, Nov 22 2023, 10:22:35) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from collections.abc import Iterable
>>> 
```

jfriel-oqc (changes requested):

src/QAT/qat/purr/compiler/builders.py
keriksson-rosenqvist: Interesting. It fails import that way on my machine and in CI. :thinking:

keriksson-rosenqvist: As seen here.
https://github.com/oqc-community/qat/actions/runs/9743002674/job/26885510222#step:5:334

keriksson-rosenqvist:
The necessary adjustments are being handled in other PRs. Dropping this in support of them.

PR #130, opened 2024-07-12
Unsafe execution engine

hamidelmaazouz:
Useful for experienced users who need to deliberately skip optimisation in order to improve runtime


hamidelmaazouz:
Closing as the use case is gonna be fulfilled in a separate more generic feature

PR #135, opened 2024-07-24
adding pypi upload

jfriel-oqc:


.github/workflows/package-publish.yml
```diff
@@ -0,0 +1,125 @@
...
+
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.x"
+    - name: Install poetry build
+      run: poetry install
+
+    - name: Build a binary wheel and a source tarball
+      run: poetry build
```
hamidelmaazouz, line 24: what about minimal qat compiler build ?

.github/workflows/package-publish.yml
```diff
@@ -0,0 +1,125 @@
...
+    - uses: actions/checkout@v4
+
+    - name: Install Poetry
+      uses: snok/install-poetry@v1
+
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: "3.x"
+    - name: Install poetry build
```
hamidelmaazouz, line 20: minor: rename to "Poetry build"

.github/workflows/package-publish.yml
```diff
@@ -0,0 +1,125 @@
...
+        GITHUB_TOKEN: ${{ github.token }}
+      # Upload to GitHub Release using the `gh` CLI.
+      # `dist/` contains the built packages, and the
+      # sigstore-produced signatures and certificates.
+      run: >-
+        gh release upload
+        '${{ github.ref_name }}' dist/**
+        --repo '${{ github.repository }}'
+    - name: Install Poetry
+      uses: snok/install-poetry@v1
```
hamidelmaazouz, line 99: this is the third "Install Poetry" step, wondering if we cache the poetry env 🤔

.github/workflows/package-publish.yml
jfriel-oqc: doing that after this, just want the core mechanism in place first

.github/workflows/package-publish.yml
jfriel-oqc: I can remove these instances actually, we only ever build once and don't want to cache anything here I don't think. Should always be a fresh build

helmaazouz (approved):

hamidelmaazouz (approved):

hamidelmaazouz:
Added https://github.com/oqc-community/qat/issues/136 for the CI failure. I've been seeing this intermittently recently. Needs investigation, but it's unrelated to this PR

jfriel-oqc:
> Added #136 for the CI failure. I've been seeing this intermittently recently. Needs investigation, but it's unrelated to this PR

I'm pretty sure it's just the runner running out of memory every now and then, but will check it out

jfriel-oqc:
failing pipe because the release already exists, I'm going to merge this

PR #137, opened 2024-07-25
Add a flag to control whether an InstructionExecutionEngine starts up…

harry-oqc:
… on instantiation. Default is True.


jfriel-oqc:
looks good, although seems like your formater is different. `peotry run format-code` should fix that

jfriel-oqc (approved):

hamidelmaazouz (approved):
Love how the commit message is used for the PR title and its continuation is in the PR description ^^

PR #138, opened 2024-07-26
CI improvements

hamidelmaazouz:
- Adds semi-automatic version control. Basically allows one to tag latest main
- Adds a conditional to publish only when a tag with a specific format is pushed to main

### DIFF
--- a/.github/workflows/cla-assistant.yaml
+++ b/.github/workflows/cla-assistant.yaml
@@ -1,10 +1,10 @@
+name: "CLA bot"
+
 on:
   issue_comment:
     types: [created]
   pull_request_target:
     types: [opened,closed,synchronize]
-
-name: "CLA bot"
 
 jobs:
   cla-acknowledgement:
--- a/.github/workflows/package-publish.yml
+++ b/.github/workflows/package-publish.yml
@@ -1,10 +1,13 @@
-name: Publish Python 🐍 distribution 📦 to PyPI and TestPyPI
+name: Publish package to PyPI and TestPyPI
 
-on: push
+on:
+  push:
+    branches: [ main ]
+    tags: [ '[0-9]+.[0-9]+.[0-9]+' ]
 
 jobs:
   build:
-    name: Build distribution 📦
+    name: Build
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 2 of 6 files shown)


.github/workflows/package-publish.yml
```diff
@@ -1,10 +1,13 @@
-name: Publish Python 🐍 distribution 📦 to PyPI and TestPyPI
+name: Publish PyPI and TestPyPI
 
-on: push
+on:
+  push:
+    branches: [ main ]
+    tags: [ '[0-9].[0-9]+.[0-9]+' ]
```
jfriel-oqc, line 6: want + on the major version, too! Let's be optimistic and assume qat will be around for more than 9 major versions!

.github/workflows/package-publish.yml
```diff
@@ -121,4 +122,4 @@ jobs:
       uses: pypa/gh-action-pypi-publish@release/v1
       with:
         password: ${{ secrets.PYPI_TEST_PUBLISH_TOKEN }}
-        repository-url: https://test.pypi.org/legacy/
\ No newline at end of file
+        repository-url: https://test.pypi.org/legacy/
```
jfriel-oqc, line 125: this can be removed, too. I can now set up the project to directly connect to github and not use tokens. Just needed the project to be created in the first instance

.github/workflows/cla-assistant.yaml
```diff
@@ -1,11 +1,11 @@
+name: "CLA bot"
+
 on:
   issue_comment:
     types: [created]
   pull_request_target:
     types: [opened,closed,synchronize]
 
-name: "CLA bot"
```
jfriel-oqc, line 7: why remove this name?

.github/workflows/version.yaml
```diff
@@ -0,0 +1,45 @@
+name: Version Control
```
jfriel-oqc, line 1: very nice :+1:

jfriel-oqc (approved):
some minor points, looks great, thanks Hamid. Approving to unblock as the comments are minor

jfriel-oqc (approved):
some minor points, looks great, thanks Hamid. Approving to unblock as the comments are minor

.github/workflows/cla-assistant.yaml
hamidelmaazouz: it's only moved to the top of file ^^ pedantic me on a Friday :)

.github/workflows/package-publish.yml
hamidelmaazouz: will leave this to another improvement sweep, we can actually create a reusable action similar to other projects

.github/workflows/cla-assistant.yaml
jfriel-oqc: somehow missed that in the diff... makes sense!

PR #140, opened 2024-07-31
CI pipeline fixes

hamidelmaazouz:
- Semi-auto version control workflow turns out to be a massive security problem.
- Changes to `main` branch should be only through PRs
- No GH action should ever be allowed to write to `main`
- Also bumping up the version manually for now as I need it for other subsequent integration tasks


chemix-lunacy:
Approving as long as people are OK with the version being bumped to 2.0 - normally you'd do this _after_ all the work has gone in and before you release. (I'm not sure if it has?)

Some minor notes on some of the comments:

* Changes to branches should mostly go through PRs but there are times when that's not viable, especially in history re-write situations.
* Github actions should absoultely be able to write to the branch they are targeting, and should be able to write to main as well. Those changes just need to be done with all the various impliciations in mind.

chemix-lunacy (approved):

bgsach (approved):
LGTM

PR #141, opened 2024-08-02
Upgrade tket to solve types-pkg-resources dependancy. 

jfriel-oqc:
Also required the upgrade of qutip, numpy, scipy and lark.

Fix tests and small RTCS change in the config from a class to a dict


helmaazouz (approved):
looks fine to me, thanks a lot Jamie

bgsach:
LGTM other than formatting failure.

hamidelmaazouz (approved):

PR #148, opened 2024-08-12
Porting Rasqal from dev to main

chemix-lunacy:
Porting Rasqal and a variety of changes from develop to main. There's quite a few things here that all intertwine.

- Replaced pyQIR with Rasqal as our QIR support. I need to add in base profile labelling to have it work with one particular test case but that is minor. I was hoping that one could be ignored really.
- All primary execution APIs now allow arguments (including for parameterized QASM files) even if it is only used for QIR.
- Interruption code has been cleaned up and pathways slightly changed. The classes can actually go at this point too, but they're harmless enough.
- Made the default hardware engine the echo if you don't pass something explicitly. RTCS is not a good default as it takes aeons to run beyond 3/4 qubits. Better to do the QASM one, but you should always have a piece of hardware.
- If you pass through no instructions anymore it just returns nothing instead of exceptions. Passing nothing _is_ actually valid in certain situations.
- Moved across minor Metrics improvements splitting enabling and initializing. Made for when you want to initialize but then enable/disable to your hearts content as execution continues.
- Allowed the fall-back Tket optimizer to process arbitrarily jagged qubits, which also then means the QASM parser now can do that as well. Right now if the QASM parser sees a register with holes it just embeds a `None` at the index where a qubit doesn't exist. Conceptually this should be fine as if you try and associate a gate to a qubit which is `None` it'll just say this qubit doesn't exist, which is accurate.

One of the bigger questions and changes here is that we're now at a point where hybrid IRs are not just QASM in disguise, so the initial flow of parse-into-instructions -> execute instructions no longer cleanly fits into that world. I believe it's time to separate out gate- and pulse-level IRs from hybrid IRs as how they function will increasingly diverge. A little bit of this can be seen here as I've made both QIR and QASM frontends have very little in common and made the parse function (almost) redundant.


chemix-lunacy:
For anyone wanting to use this to test there are a few things to know:

1. Routing is currently disabled, you can enable it by searching for `# TODO: Routing disabled for now.`. It just needs another pass to make sure what it's doing is correct.
2. It consumes unrestricted QIR/LLVM IR but any instructions outside the QIR base profile is considered 'experimental'. This means there is no guarantee they are implemented yet, or that every permutation of them have been tested. Most of the common instructions are available and work, but there may untested scenarios in which bugs may occur.
3. There is a default limit on how many steps the executor can take. By default this is 2500, which is more than you would expect, but it will cut-off if the QIR would execute too much.

**If you want to see precisely what Rasqal is doing use its debug logging.** 

First call `initialize_logger` with a path where you want the log file to be written to. Then use the `RasqalRunner` object in `qir.py` to activate the varied logging mechanisms:

`trace_graphs()` Prints out the graph the QIR gets parsed into just before it gets executed. 
`trace_projections()` Prints out precisely what the projections are doing, what circuit they are executing, when they hit a cache etc.
`trace_runtime()` Prints out each precise step the runtime takes including current context, active variables and scoping information.

Each of these emit a huge amount of information but will give you a real view of what is happening internally. They use a visual view to represent their data which may not be immediately obvious what it represents and right now does not have a document on how to read them. Just send me a quick message with fragments of the output if you need any assistance.

bgsach:
Not considering further for now

PR #153, opened 2024-08-19
Implement return values for startup/shutdown methods of LiveDeviceEngine

harry-oqc:
Currently, startup and dosconnect calls for LiveDeviceEngine return None. Changed to return overall instrument connectivity.

### DIFF
--- a/src/qat/purr/backends/live.py
+++ b/src/qat/purr/backends/live.py
@@ -230,6 +230,8 @@ class LiveDeviceEngine(QuantumExecutionEngine):
         return all(is_connected)
 
     def shutdown(self):
+        # Store the instrument connection statuses.
+        # All instruments should be disconnected (represented as a False entry)
         is_connected = []
         for instrument in self.model.instruments.values():
             is_connected.append(instrument.close())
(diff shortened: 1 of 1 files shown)


src/qat/purr/backends/live.py
```diff
@@ -213,23 +213,30 @@ def __exit__(self, exc, value, tb):
...
         for physical_channel in self.model.physical_channels:
             if isinstance(physical_channel, ControlHardwareChannel):
                 for dc_bias_ch in physical_channel.dcbiaschannel_pair.values():
-                    dc_bias_ch.connect_to_instrument()
+                    is_connected.append(dc_bias_ch.connect_to_instrument())
+
+        return all(is_connected)
 
     def shutdown(self):
+        is_connected = []
```
jfriel-oqc, line 233: this is a bit confusing in terms of what true/false means here. Can we just get a quick comment to be very explicit on what true means and what you expect to return when everything is disconnected correctly.

harry-oqc: Great - I will see to it

jfriel-oqc (approved):

PR #154, opened 2024-08-20
Qblox debug - Fix file name datetime format for windows

hamidelmaazouz:


src/qat/purr/backends/qblox/device.py
```diff
@@ -223,7 +223,7 @@ def install(self, package: QbloxPackage):
 
         sequence = asdict(package.sequence)
         if self.dump_sequence:
-            filename = f"schedules/sequence_{module.slot_idx}_{sequencer.seq_idx}_@_{datetime.utcnow()}.json"
+            filename = f"schedules/sequence_{module.slot_idx}_{sequencer.seq_idx}_@_{datetime.utcnow().strftime('%m-%d-%Y_%H%M%S')}.json"
```
bgsach, line 226: Do we need to worry about backwards compatibility here?

hamidelmaazouz: nah this is only useful for debugging, also no scheme had been in place for post processing of the dumped files, so all good

jfriel-oqc (approved):

PR #155, opened 2024-08-28
Alter startup/shutdown engine connection metrics. Add engine_is_connected metrics

harry-oqc:


harry-oqc:
> what's the purpose of this change? my understanding was that the .connect() method should return .is_connected anyway

Unfortunately this is not the case, down to all the method overriding in whisqrs through the inheritance that occurs there. When we experimented without this change, we only have None returned in all cases.

jfriel-oqc (approved):

PR #156, opened 2024-08-29
Qblox improvements

hamidelmaazouz:
- Manipulate digital offset when the pulse envelope is trivial such as a square
- Make code reusable between a drive pulse and a measure pulse (+ acquire)
- Sanitisation pulse width

### DIFF
--- a/src/qat/purr/backends/qblox/codegen.py
+++ b/src/qat/purr/backends/qblox/codegen.py
@@ -328,6 +328,11 @@ class QbloxContext:
             self.sequence_builder.wait(remainder)
 
     def _evaluate_waveform(self, waveform: Waveform, target: PulseChannel):
+        """
+        The waveform is evaluated as a 1d complex array. In QBlox, the Real and Imag parts of the pulse
+        represent the digital offset on the AWG. They both must be within range [-1, 1].
+        """
+
         samples = int(calculate_duration(waveform, return_samples=True))
 
         dt = target.sample_time
--- a/src/qat/purr/backends/qblox/codegen.py
+++ b/src/qat/purr/backends/qblox/codegen.py
@@ -333,13 +333,24 @@ class QbloxContext:
         represent the digital offset on the AWG. They both must be within range [-1, 1].
         """
 
-        samples = int(calculate_duration(waveform, return_samples=True))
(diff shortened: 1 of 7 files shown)


src/qat/purr/backends/qblox/codegen.py
```diff
@@ -331,34 +346,41 @@ def _process_waveform(self, waveform: Waveform, target: PulseChannel):
         pulse *= scale
         pulse += target.bias
 
-        if 2 * pulse.size > self._wf_memory:  # for both I and Q
+        if ((pulse.real < -1) | (pulse.real > 1)).any():
```
jfriel-oqc, line 349: do we not need this to be the magnitude of the pulse? If not can we add a comment to be explicit that we're testing components, not magnitude here

src/qat/purr/backends/qblox/codegen.py
```diff
@@ -374,36 +396,119 @@ def delay(self, inst: Delay):
...
+            wf_name = type(waveform).__name__
+            if isinstance(waveform, Pulse):
+                wf_name = waveform.shape.name
+
+            i_index, q_index = self._register_waveform(wf_name, pulse)
+            self.sequence_builder.play(i_index, q_index, num_samples)
+
         self._duration = self._duration + waveform.duration
 
     def measure_acquire(
```
jfriel-oqc, line 441: this is a very large method with nested functions. Please refactor to be cleaner and remove nested functions. This gives very unclear behaviour and makes granular testing very difficult

src/qat/purr/backends/qblox/device.py
```diff
@@ -183,6 +186,27 @@ def _prepare_config(self, package: QbloxPackage, sequencer: Sequencer):
...
         return qblox_config
 
+    def reset(self, hard=False):
+        """
+        TODO - Needs a rethink when daisy-chaining clusters
+        """
+
+        self._resources.clear()
+
+        if hard:
```
jfriel-oqc, line 196: can we add a comment in the doc string on the difference between a hard reset and a soft reset

jfriel-oqc (approved):
nothing major, looks great! Just a few comments, approving to unblock. Only one that I would personally defo want to see addressed is simplifying the large method and removing nested methods

PR #160, opened 2024-09-01
Qblox auto mixer calibration

hamidelmaazouz:
- Minor refactoring
- More convenient milti-level reset
- Auto mixer cal

### DIFF
--- a/src/qat/purr/backends/qblox/config.py
+++ b/src/qat/purr/backends/qblox/config.py
@@ -170,8 +170,8 @@ class QbloxConfig:
 
 
 class QbloxConfigHelper(ABC):
-    def __init__(self, config: QbloxConfig):
-        self.config = config
+    def __init__(self, config: QbloxConfig = None):
+        self.config = config or QbloxConfig()
 
     def configure(self, module: Module, sequencer: Sequencer):
         self.configure_module(module, self.config.module)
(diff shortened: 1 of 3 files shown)


src/qat/purr/backends/qblox/device.py
```diff
@@ -120,12 +129,12 @@ def __init__(
...
         name: str = None,
         address: str = None,
-        cfg: Dict = None,
+        dummy_cfg: Dict = None,
     ):
         super().__init__(id_=dev_id or os.environ.get("QBLOX_DEV_ID"))
         self.name = name or os.environ.get("QBLOX_DEV_NAME")
         self.address = address or os.environ.get("QBLOX_DEV_IP")
-        self.cfg = cfg
+        self.dummy_cfg = dummy_cfg
```
keriksson-rosenqvist, line 137: Why `dummy`?

src/qat/purr/backends/qblox/config.py
```diff
@@ -170,7 +170,7 @@ class QbloxConfig:
 
 
 class QbloxConfigHelper(ABC):
-    def __init__(self, config: QbloxConfig):
+    def __init__(self, config=QbloxConfig()):
```
keriksson-rosenqvist, line 173: Minor but why not keep the type hint as is but optional and do:
```
def __init__(self, config: QbloxConfig = None):
    self.config = config or QbloxConfig()
```

src/qat/purr/backends/qblox/device.py
hamidelmaazouz: config for dummy qblox cluster, it's a "mock" cluster qblox provides for testing

bgsach (approved):
Talked through in person.

PR #163, opened 2024-09-02
Give ID to setup python step

alillistone-OQC:
The "Load cached venv" step was unable to append the python version to the name of the python venv cache. This was causing sporadic failures of the pipelines. This change will give the step "Set up Python 3.x" and ID "setup-python" so it can correctly fetch the cache.

### DIFF
--- a/.github/workflows/package-build.yaml
+++ b/.github/workflows/package-build.yaml
@@ -30,9 +30,10 @@ jobs:
 
       - name: Set up Python 3.x
         uses: actions/setup-python@v5
+        id: setup-python
         with:
           python-version: ${{ matrix.python }}
-          cache: 'pip'
+          cache: 'poetry'
 
       - name: Install Poetry
         uses: snok/install-poetry@v1
--- a/.github/workflows/package-build.yaml
+++ b/.github/workflows/package-build.yaml
@@ -75,10 +75,10 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Set up Python 3.x
+        id: setup-python
         uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python }}
-          cache: 'pip'
 
       - name: Install Poetry
         uses: snok/install-poetry@v1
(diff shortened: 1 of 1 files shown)


alillistone-OQC:

.github/workflows/package-build.yaml
```diff
@@ -30,9 +30,10 @@ jobs:
 
       - name: Set up Python 3.x
         uses: actions/setup-python@v5
+        id: setup-python
         with:
           python-version: ${{ matrix.python }}
-          cache: 'pip'
+          cache: 'poetry'
```
keriksson-rosenqvist: Not convinced this should be poetry as it's not installed yet.

.github/workflows/package-build.yaml
```diff
@@ -30,9 +30,10 @@ jobs:
 
       - name: Set up Python 3.x
         uses: actions/setup-python@v5
+        id: setup-python
```
keriksson-rosenqvist, line 33: Probably need this ID in the other setup python step(s) as well. 
https://github.com/oqc-community/qat/pull/163/files#diff-39100e6539ba9a86c5830cf257ebdfdef8797d743f9787ec1be460f1e8dc941aR79

keriksson-rosenqvist (changes requested):

.github/workflows/package-build.yaml
```diff
@@ -30,9 +30,9 @@ jobs:
...
         with:
           python-version: ${{ matrix.python }}
-          cache: 'pip'
```
bgsach, line 35: Why did you also drop this?

alillistone-OQC: I removed this as it doesn't do anything. We use poetry to install dependencies which installed the dependencies in a different directory to pip.

alillistone-OQC: https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages

keriksson-rosenqvist (approved):
LGTM

bgsach (commented):
LGTM

PR #172, opened 2024-09-09
Bugfix if last instruction was measure

willh727-OQC:
Fixes bug with measure is the last instruction which returns tuple from `FluidBuilderWrapper`. Checks if iter has a builder and calls the builders instructions instead. This bug was found in the test case:
```
builder = (get_builder(model)
                   .X(drive_channel, np.pi / 2.0)
                   .measure(qubit))
results, _ = execute_instructions(engine, builder)
```
Which only returned a singular acquire. Instead of processing the drive pulse and measure.

### DIFF
--- a/src/qat/purr/compiler/runtime.py
+++ b/src/qat/purr/compiler/runtime.py
@@ -193,7 +193,7 @@ class QuantumRuntime(MetricsMixin):
         if self.engine is None:
             raise ValueError("No execution engine available.")
 
-        if isinstance(instructions, InstructionBuilder):
+        if isinstance(instructions, InstructionBuilder) or any([i for i in instructions if isinstance(i, InstructionBuilder)]):
             instructions = instructions.instructions
 
         if instructions is None or not any(instructions):
--- a/src/qat/purr/compiler/runtime.py
+++ b/src/qat/purr/compiler/runtime.py
@@ -5,6 +5,7 @@ from numbers import Number
 from typing import List, Optional, TypeVar, Union
 
 import numpy
+from purr.compiler.builders import FluidBuilderWrapper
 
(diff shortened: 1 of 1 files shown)


willh727-OQC:
Happy for suggestions of a better fix, not sure if I fully understanding how the FluidBuilderWrapper works

src/qat/purr/compiler/runtime.py
```diff
@@ -193,7 +193,7 @@ def _common_execute(
         if self.engine is None:
             raise ValueError("No execution engine available.")
 
-        if isinstance(instructions, InstructionBuilder):
+        if isinstance(instructions, InstructionBuilder) or any([i for i in instructions if isinstance(i, InstructionBuilder)]):
```
hamidelmaazouz, line 196: directly `... or isinstance(instructions, FluidInstructionBuilder)`, otherwise you're not guaranteed it'll have a `instructions` field

hamidelmaazouz (changes requested):

src/qat/purr/compiler/runtime.py
willh727-OQC: Ben just mentioned that to me as well! Fixed

hamidelmaazouz:
[Failure here](https://github.com/oqc-community/qat/actions/runs/10777640052/job/29887171055?pr=172) is irrelevant. You're probably opening the PR from your own private fork. The Doc step tries to checkout the same branch name (on the base repo and fails)
- One solution is reopen the PR but from the base repo. Currently not ready as you need write access
- Push the same commits under the same branch name to the base repo, and rerun the Doc step

willh727-OQC:
Also one of the unit tests fails, the involves generating random numbers

PR #173, opened 2024-09-10
fix: resolve flaky noise simulation test

lcauser-oqc:
The test checks the functionality to mitigate the readout error using a linear model by running a noisy simulator. It checks that the returned mitigated results fall within a certain tolerance of the expected result.

The problem is that the results are statistical, and by chance can fall outside of this tolerance (it happened about 1% of the time in my checks...).

The solution here isn't a perfect one. As the test is implemented, it would be nearly impossible to reduce the chance to fail to zero in a meaningful way. I have just added some repeats in the experiment to reduce the variance of the (averaged) deviation with the expected result, which should make it pass more reliably. Maybe in the future we should reconsider the purpose of the test, as a test that relies on randomness is not ideal.

### DIFF
--- a/tests/qat/test_readout_mitigation.py
+++ b/tests/qat/test_readout_mitigation.py
@@ -328,6 +328,8 @@ class TestOnNoisySimulator:
         q0_ro_fidelity_1=0.7,
         q1_ro_fidelity_0=0.9,
         q1_ro_fidelity_1=0.6,
+        threshold=0.051,
+        repeats=5
     ):
         assert len(bitstring) == 2
         circuit = QuantumCircuit(2, 2)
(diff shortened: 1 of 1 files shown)


lcauser-oqc:

hamidelmaazouz:
could we just seed the random generator ?

lcauser-oqc:
> could we just seed the random generator ?

This was my original thought - I mostly opposed the idea because I think fixing the output of a random simulation should be replaced with avoiding a test with random outcomes. But this solution I've done here doesn't deal with that either :') For now I will fix the seed (but maybe we should think about the purpose of the test later). Thanks!

tests/qat/test_readout_mitigation.py
```diff
@@ -328,8 +328,14 @@ def test_prepare_bitstring_and_measure(
         q0_ro_fidelity_1=0.7,
         q1_ro_fidelity_0=0.9,
         q1_ro_fidelity_1=0.6,
+        threshold=0.051,
     ):
         assert len(bitstring) == 2
+        # Fix the seed for the experiment
+        np.random.seed(0)
+        seed(0)
```
hamidelmaazouz, line 336: why are we seeding using both the python core library and numpy ? isn't numpy enough ?
For repurposing the test, I suggest add a follow up ticket and involve Bryn as he originally put up the random sim test

hamidelmaazouz (commented):
looks nice, minor tweak

tests/qat/test_readout_mitigation.py
lcauser-oqc: > why are we seeding using both the python core library and numpy ? isn't numpy enough ? For repurposing the test, I suggest add a follow up ticket and involve Bryn as he originally put up the random sim test

RealtimeChipSimEngine uses random() from the core library - had to seed both to have a certain result

hamidelmaazouz (approved):

Issue #76, opened 2024-03-08
Set system shot limit


keriksson-rosenqvist-oqc:
**Add setting/configuration option to set maximum shot limit.**

Adding a system setting that can be used to limit the maximum number of shots any single job can have. Though high numbers of shots could be split into batches the system owner may want to set a limit to limit the runtime and protect against memory issues.

**Specific details of what you would like to see**

The system owner can define a system level value to define the maximum number of shots that any single task can have.

motazmuhammad:
@keriksson-rosenqvist-oqc I am a bit confused about this one, is it still unimplemented? it seems to be implemented but not closed yet. Is there a reason for that?

PR #175, opened 2024-09-10
COMPILER-8 Limit system settings.

daria-oqc:
This PR implements an error throw if the total number of shots exceeds 100000.


keriksson-rosenqvist (changes requested):

src/qat/purr/compiler/config.py
```diff
@@ -248,6 +248,12 @@ def create_from_json(cls, json: str):
     def validate(self, hardware):
         from qat.purr.compiler.hardware_models import QuantumHardwareModel
 
+        # Throw error if the number of errors exceeds the limit on qcaas.
+        if self.repeats and self.repeats > 100000:
```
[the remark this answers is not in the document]
keriksson-rosenqvist: See related issue https://github.com/oqc-community/qat/issues/76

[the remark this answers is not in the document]
bgsach: As discussed in person, we will use pydantic for qat level config.

src/qat/purr/qatconfig.py
```diff
@@ -0,0 +1,14 @@
...
+    """
+    Full settings for a single job. All values are defaulted on initialization.
+    """
+
+    model_config = SettingsConfigDict(env_prefix="QAT_")
+    MAX_REPEATS_LIMIT: int = 100_000
+    """Max number of repeats / shots to be performed in a single job."""
+
+
+qatconfig = QatConfig()
```
bgsach, line 14: I think we probably want to put

``` from qat.purr.qatconfig import qatconfig``` in qat/__init__.py so people can do `from qat import qatconfig` 

?

Comment (daria-oqc) [reply src/qat/purr/qatconfig.py:14]: Indeed, but that would be more of a general comment as there are some other functionalities which should be imported within `__init__` as well to allow for easier imports?

File: tests/qat/test_config.py [hunk 3]
```diff
@@ -105,6 +107,24 @@ def test_config_repeats(self):
...
         assert first_conf.repetition_period == second_conf.repetition_period
 
+    @pytest.mark.parametrize(
+        "repeats, possible_error_throw",
+        [
+            (64_000, does_not_raise()),
+            (100_000, does_not_raise()),
+            (100_001, pytest.raises(ValueError, match="exceeds the maximum amount of")),
+            (1_000_000, pytest.raises(ValueError, match="exceeds the maximum amount of")),
+            (0, does_not_raise()),
```
keriksson-rosenqvist, line 117: Super minor.
Could these be ordered `None` - highest?
Also, suggest using [PyTest ids](https://docs.pytest.org/en/stable/example/parametrize.html#different-options-for-test-ids) to make the output more informative.

Some examples: 
* https://github.com/oqc-community/qat/blob/7df92aed503ae4a7f6a11ec22965c3fa831a9b99/tests/qat/test_tket.py#L22
* https://github.com/oqc-community/qat/blob/7df92aed503ae4a7f6a11ec22965c3fa831a9b99/tests/qat/test_instructions.py#L264
* https://github.com/oqc-community/qat/blob/7df92aed503ae4a7f6a11ec22965c3fa831a9b99/tests/qat/test_instructions.py#L312

keriksson-rosenqvist (commented):
Looks generally good. Could we get a test/example where the value is adjusted by the environment? :thinking:

keriksson-rosenqvist (approved):

PR #177, opened 2024-09-11
Bugfix if last instruction was measure [getting around pipline]

willh727-OQC:
same as https://github.com/oqc-community/qat/pull/172 but same name as @hamidelmaazouz mentioned


hamidelmaazouz (approved):

hamidelmaazouz (approved):
Actually, no, add

PR #178, opened 2024-09-11
Bring back cached properties

hamidelmaazouz:


daria-oqc (approved):
👍🏻

lcauser-oqc (approved):
looks good to me

bread-oqc:
This is a breaking change for dependent components. To avoid the breaking change, we could cache and return a lambda so that the expression '*.full_id' is still callable but this is grosssss.

I reckon this is the 'correct' solution.

bgsach:
There are concerns from Kajsa about mutability here. We should pause and have a think.

hamidelmaazouz:
> There are concerns from Kajsa about mutability here. We should pause and have a think.

@keriksson-rosenqvist could you share where the mutability requirement comes from ?

PR #179, opened 2024-09-11
COMPILER-69 Add log info for each (dis)connected instrument.

daria-oqc:


hamidelmaazouz (approved):
Change is fine but would love to see some context in the PR desc

PR #182, opened 2024-09-13
ak_wh/hotfix_weights

willh727-OQC:
Fixing Acquire filter to handle floating points in a consistent way, to how its produced in calculating durations


hamidelmaazouz:
@willh727-OQC consistency issues can be fixed by running `poetry run format-code`

src/qat/purr/compiler/instructions.py
```diff
@@ -348,10 +349,11 @@ def _check_filter(self, filter):
...
                 if isinstance(target, PulseChannel):
                     dt = target.physical_channel.sample_time
-                    if not np.isclose(filter.duration, dt * (self.time // dt), atol=1e-12):
+                    no_samples_from_time = math.floor(round(self.time / dt))
+                    if not np.isclose(filter.duration, dt * no_samples_from_time, atol=1e-12):
                         raise ValueError(
                             f"Filter duration '{filter.duration}' must be equal to Acquire "
-                            f"duration '{self.time}'."
+                            f"duration '{dt * no_samples_from_time}' which was rounded."
                         )
```
hamidelmaazouz, line 357: similar processing as `XEngine.calculate_duration()` https://github.com/oqc-community/qat/blob/main/src/qat/purr/compiler/execution.py#L401
It's now an instance method, you could rip it off, make it a module utility function in `instructions.py` module and use it for your case:

```
calculate_duration(self)
```

willh727-OQC: Yeah good idea, I'll also replace it in calcaulte_duration so its done it one place

elskysr:
recheck

src/qat/purr/compiler/instructions.py
```diff
@@ -347,11 +349,14 @@ def _check_filter(self, filter):
...
-                    if not np.isclose(filter.duration, dt * (self.time // dt), atol=1e-12):
+                    if not np.isclose(
+                        filter.duration,
+                        calculate_duration(self, return_samples=False),
+                        atol=1e-9,
+                    ):
                         raise ValueError(
                             f"Filter duration '{filter.duration}' must be equal to Acquire "
-                            f"duration '{self.time}'."
+                            f"duration '{calculate_duration(self, return_samples=False)}'."
```
hamidelmaazouz, line 359: put this in a variable `duration_nanos = calculate_duration(self, return_samples=False)` and reuse here

src/qat/purr/compiler/instructions.py
```diff
@@ -683,3 +688,45 @@ def __init__(self, variable: str, res_processing: InlineResultsProcessing):
 
     def __repr__(self):
         return f"{self.variable}: {str(self.results_processing.name)}"
+
+
+def remove_floating_point(x):
```
hamidelmaazouz, line 693: inline this function please, it's too short and I don't think it's used anywhere else

tests/qat/test_instructions.py
```diff
@@ -171,23 +172,27 @@ def test_012_entanglement(self):
             qubit2: {qubit0, qubit1, qubit2},
         }
 
-    def test_acquire_filter(self):
+    @pytest.mark.parametrize(
+        "acquire_width",
+        np.linspace(1e-6, 6e-6, 10),
```
hamidelmaazouz, line 177: consider edge cases: negative, 0, some special positive value, ...

hamidelmaazouz (approved):
Love the work 😃  a few comments but I'm already happy. Also approving to unblock but please carry out the last changes I pointed to. Thanks for the fix @willh727-OQC

willh727-OQC:
> Love the work 😃 a few comments but I'm already happy. Also approving to unblock but please carry out the last changes I pointed to. Thanks for the fix @willh727-OQC

Don't worry I can't merge it anyway!

src/qat/purr/compiler/instructions.py
willh727-OQC: I get your point! But I'm planning to use it in Whisqrs for ArbSeq! Just to make the removal of floating points consistent and the same as QAT.

hamidelmaazouz: the only thing worth sharing here is the rounding precision, which can be set properly via compiler config

cc/ @bgsach

hamidelmaazouz: It's okay for now, assume this function is gonna disappear soon, and that you need to pick the rounding precision from the config instead, I don't want to see apis leaking this way between qat and internal dependencies

hamidelmaazouz (approved):

src/qat/purr/compiler/instructions.py
willh727-OQC: Alright good to know theres a better solution going forward.

hamidelmaazouz (approved):

PR #183, opened 2024-09-16
HW-feasible settings for pulse durations.

daria-oqc:
- Throw warning if the gate duration from the hardware set-up is not within the granularity of the HW.
- Set default X(pi/2) duration conform to granularity (which changes the timeline of a few tests as well).

See COMPILER-90 ticket for more info.


hamidelmaazouz (changes requested):
it's a ctrl hw concern, should not be associated with a qubit, please move to physical channel.

Also add more tests

hamidelmaazouz:
also checkout `PhysicalChannel.block_size`

bgsach (approved):
LGTM subject to hamid's comments

tests/qat/test_quantum_backend.py
```diff
@@ -1,5 +1,6 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Oxford Quantum Circuits Ltd
+import logging
```
hamidelmaazouz, line 3: rogue import, please use QAT logger

tests/qat/test_quantum_backend.py
```diff
@@ -362,6 +366,54 @@ def test_setup_hold_measure_pulse(self):
...
+    def test_pulse_validation_grid_time_throws_warning_for_sweep(self, caplog):
+        hw = get_test_model()
+        qubit = hw.get_qubit(0)
+        drive_channel = qubit.get_drive_channel()
+        engine = get_test_execution_engine(hw)
+
+        time = np.linspace(
+            4e-09, 1.6e-8, 11
+        )  # Contains non-hardware-feasible pulse widths.
+        print(time)
```
hamidelmaazouz, line 405: log here instead of print ^^

tests/qat/test_quantum_backend.py
```diff
@@ -362,6 +366,54 @@ def test_setup_hold_measure_pulse(self):
...
+        hw = get_test_model()
+        qubit = hw.get_qubit(0)
+        drive_channel = qubit.get_drive_channel()
+
+        builder = get_builder(hw).pulse(drive_channel, PulseShapeType.SQUARE, width=width)
+
+        engine = get_test_execution_engine(hw)
+
+        # Capture if warnings are sent to logger.
+        with caplog.at_level(logging.WARNING):
```
hamidelmaazouz, line 389: oh you're using it here to catch log events

src/qat/purr/backends/live.py
```diff
@@ -82,8 +82,23 @@ def apply_setup_to_hardware(
         qubit = Qubit(primary_index, resonator, ch1)
         qubit.create_pulse_channel(ChannelType.drive, frequency=5.5e9)
 
+        grid_time = qubit.physical_channel.grid_time
```
hamidelmaazouz, line 85: can you specify which time units we're dealing with here ?
usually it's `name_nanos` or `name_seconds`
integer arithmetic suggests the unit is nanos but just to be explicit

src/qat/purr/backends/live.py
```diff
@@ -82,8 +82,23 @@ def apply_setup_to_hardware(
...
+        if not pulse_hw_x_pi_2_width % grid_time:
```
hamidelmaazouz, line 86: prefer `if a % b > 0` instead

src/qat/purr/backends/live.py
```diff
@@ -82,8 +82,23 @@ def apply_setup_to_hardware(
...
+            log.warning(
+                f"Non-hardware-feasible pulse width detected in hardware setup: pulse_hw_x_pi_2_width is not a multiple of the pulse resolution, which is {np.round(grid_time * 1e09, 3)} ns."
```
hamidelmaazouz, line 88: too long a message, say simply "Pulse width must/should be a multiple of {val}"
also why are you logging a warning ? If it's a hard requirement then please raise an error (you could think about rounding up) (depending on semantics)

src/qat/purr/backends/live.py
```diff
@@ -82,8 +82,23 @@ def apply_setup_to_hardware(
...
+        if not pulse_hw_zx_pi_4_width % grid_time:
+            log.warning(
+                f"Non-hardware-feasible pulse width detected in hardware setup: pulse_hw_zx_pi_4_width is not a multiple of the pulse resolution, which is {np.round(grid_time * 1e09, 3)} ns."
+            )
         qubit.pulse_hw_zx_pi_4.update({"width": pulse_hw_zx_pi_4_width})
+
+        if not pulse_measure_width % grid_time:
+            log.warning(
+                f"Non-hardware-feasible pulse width detected in hardware setup: pulse_measure_width is not a multiple of the pulse resolution, which is {np.round(grid_time * 1e09, 3)} ns."
+            )
```
hamidelmaazouz, line 101: same as above

src/qat/purr/backends/live.py
```diff
@@ -82,8 +82,23 @@ def apply_setup_to_hardware(
...
+        if not pulse_hw_x_pi_2_width % grid_time:
+            log.warning(
+                f"Non-hardware-feasible pulse width detected in hardware setup: pulse_hw_x_pi_2_width is not a multiple of the pulse resolution, which is {np.round(grid_time * 1e09, 3)} ns."
+            )
         qubit.pulse_hw_x_pi_2.update({"width": pulse_hw_x_pi_2_width})
+
+        if not pulse_hw_zx_pi_4_width % grid_time:
+            log.warning(
+                f"Non-hardware-feasible pulse width detected in hardware setup: pulse_hw_zx_pi_4_width is not a multiple of the pulse resolution, which is {np.round(grid_time * 1e09, 3)} ns."
+            )
```
hamidelmaazouz, line 95: same as above

src/qat/purr/backends/live.py
hamidelmaazouz: also consider grouping the lot in a loop:

```
for width in [pulse_hw_x_pi_2_width, pulse_hw_zx_pi_4_width, pulse_measure_width]:
    if not width % grid_time:
        # round up, log, or raise
```

src/qat/purr/compiler/devices.py
```diff
@@ -315,6 +315,7 @@ def __init__(
         acquire_allowed: bool = False,
         pulse_channel_min_frequency: float = 0.0,
         pulse_channel_max_frequency: float = np.inf,
+        grid_time: float = 4e-09,  # 4 ns
```
hamidelmaazouz, line 318: ah, it's seconds, I'd be very careful with arithmetic operations on non integer types. Here it's fine, keep it as seconds, when you use the field, make the necessary conversions

hamidelmaazouz: in general this verification feels like it should be a constant rather than an instance field. Ideally you'd have some enum `Constants` and use it like `Constants.GRID_TIME`.

I also think it should not belong here and should rather be implemented internally as it's backend-specific.

src/qat/purr/compiler/devices.py
```diff
@@ -699,6 +701,7 @@ def __init__(
         coupled_qubits: List[Qubit] = None,
         drive_amp: float = 1.0,
         id_=None,
+        **kwargs,
```
hamidelmaazouz, line 704: can't find any usages of this parameter

src/qat/purr/compiler/devices.py
```diff
@@ -713,7 +716,7 @@ def __init__(
 
         self.pulse_hw_x_pi_2 = {
             "shape": PulseShapeType.GAUSSIAN,
-            "width": 100e-9,
+            "width": 9.6e-08,
```
hamidelmaazouz, line 719: why is this change ?

src/qat/purr/compiler/execution.py
```diff
@@ -243,27 +244,30 @@ def validate(self, instructions: List[Instruction]):
...
+                    values = np.array(
+                        next(
+                            iter(
+                                [
+                                    sw.variables[duration.name]
+                                    for sw in instructions
+                                    if isinstance(sw, Sweep)
+                                    and duration.name in sw.variables.keys()
+                                ]
+                            )
```
hamidelmaazouz, line 259: too bulky, compute the associated sweep first and then get the variable value from it

src/qat/purr/compiler/execution.py
```diff
@@ -243,27 +244,30 @@ def validate(self, instructions: List[Instruction]):
...
-                        )
+
+                if np.any(values > MaxPulseLength):
+                    raise ValueError(
+                        f"Max Waveform width is {MaxPulseLength} s, given: {values} s."
+                    )
+                elif np.any(values % grid_time):
+                    log.warning(
+                        f"Non-hardware-feasible pulse detected: {inst} does not have a pulse width that is a multiple of the pulse resolution ({np.round(grid_time * 1e09, 3)} ns)."
+                    )
```
hamidelmaazouz, line 270: same as comments on on the `live.py` module

tests/qat/test_quantum_backend.py
hamidelmaazouz: if you're raising an error then this needs to go away

hamidelmaazouz (changes requested):

tests/qat/test_quantum_backend.py
daria-oqc: Oh that one was during testing that I forgot to remove afterwards :) thanks!

src/qat/purr/compiler/devices.py
daria-oqc: Moved internally.

src/qat/purr/compiler/devices.py
daria-oqc: To respect with HW specific granularity. Since I will be moving this feature internally, this change is no longer necessary.

daria-oqc:
Closed this one due to move internally, as discussed with @hamidelmaazouz .

PR #184, opened 2024-09-16
Performance improvement: create_duration_timeline

lcauser-oqc:
Changes:
- Repeated calculations brought outside of loops to give performance improvements.
- Additional tests to verify that `create_duration_timeline` behaves as expected.


bgsach (commented):
Looks good. Are these changes well covered by existing tests?

lcauser-oqc:
> Looks good. Are these changes well covered by existing tests?

The function should be tested anytime a test executes a circuit (maybe with the exception of the qiskit hardware model). However, I'm not certain if these will test it functions as expected of if it just executes the code, so I've added a test with a simple circuit to test that it works as we expect

src/qat/purr/compiler/execution.py
```diff
@@ -453,6 +453,14 @@ def create_duration_timeline(self, instructions: List[QuantumInstruction]):
         total_durations: Dict[PulseChannel, int] = dict()
 
         for instruction in instructions:
+            # calculate the total durations for each component
+            if isinstance(instruction, Synchronize):
+                current_durations = [
+                    total_durations.setdefault(qt, 0) for qt in instruction.quantum_targets
+                ]
+                longest_length = max(current_durations, default=0.0)
+                ctr = 0  # count to eliminate dict search later
```
hamidelmaazouz, line 462: how do you explain the speedup by just using `ctr` as index ?

hamidelmaazouz (changes requested):

src/qat/purr/compiler/execution.py
hamidelmaazouz: > it might be redundant later if we can cache the full id?
correct. keys in the original dictionary (`total_durations`) are pulse channels. When you read a key, the pulse channel gets hashed which gets `full_id` to be called. So what you're doing here is circumvent hashing pulse channels to hashing just integers, which becomes trivial.

The overall improvement here shows one beneficial side of caching `full_id`.

lcauser-oqc: Yeah good point! I guess for readability, it would be better to keep `current_durations` as a dict with the hope that these issues are addressed later with caching.

Note that it is still worth calculating `longest_length` outside of the loop (in practice, even with the dict and caching, I see some speed-up by doing this).

daria-oqc (approved):
👍🏻

keriksson-rosenqvist (approved):
LGTM

tests/qat/test_quantum_backend.py
```diff
@@ -714,6 +714,163 @@ def test_create_duration_timeline_mapping(self):
             *[PostProcessing] * 4,
         ]
 
+    @pytest.mark.parametrize(
+        "hw", [get_test_model(), get_default_echo_hardware(), get_default_RTCS_hardware()]
+    )
+    def test_duration_timeline_times(self, hw):
+        # Tests that the creation of a duration timeline with a two-qubit circuit
+        # gives a timeline where each pulse is concurrent.
```
hamidelmaazouz, line 722: what's concurrent here exactly ? do you mean the x pi and x pi/2 ?

tests/qat/test_quantum_backend.py
```diff
@@ -714,6 +714,163 @@ def test_create_duration_timeline_mapping(self):
...
+            .measure_mean_z(q2)
+        )
+        res1 = engine.create_duration_timeline(builder.instructions)
+
+        # Check the times match up
+        for _, val in res1.items():
+            end = 0
+            for pos in val:
+                assert end == pos.start
+                end = pos.end
```
hamidelmaazouz, line 743: this for loop only tests chaining of start, end, start, end on a single target.
Could you add more checks that contrasts a timeline of a target against another ?

tests/qat/test_quantum_backend.py
```diff
@@ -714,6 +714,163 @@ def test_create_duration_timeline_mapping(self):
...
+
+    @pytest.mark.parametrize(
+        "hw", [get_test_model(), get_default_echo_hardware(), get_default_RTCS_hardware()]
+    )
+    def test_duration_timeline_sync(self, hw):
+        # Tests that a redundant sync has no effect on the circuit.
```
hamidelmaazouz, line 749: makes sense as an ECR starts with a Sync. I would also try out another scenario where the added Sync would be on top of naturally/coincidentally synced timelines:

```
target1 : 1000 ns tune              + P
target2 : 500 ns tune + 500 ns tune + P
```

Everything that comes before P happens to align naturally, so adding a sync there would not practically affect timelines

tests/qat/test_quantum_backend.py
```diff
@@ -714,6 +714,163 @@ def test_create_duration_timeline_mapping(self):
...
+                for inst in res1[key]
+                if (not isinstance(inst.instruction, Delay) and inst.start != inst.end)
+            ]
+            ends2 = [
+                inst.start
+                for inst in res2[key]
+                if (not isinstance(inst.instruction, Delay) and inst.start != inst.end)
+            ]
+            assert starts1 == starts2
+            assert ends1 == ends2
```
hamidelmaazouz, line 799: how the timeline contrast is tested here is fine but could improve:

look for a specific instruction other than a delay in both targets and make sure they start at the same time.
(measure/acquire would be a good example)
(and of course keep the delay case)

tests/qat/test_quantum_backend.py
```diff
@@ -714,6 +714,163 @@ def test_create_duration_timeline_mapping(self):
...
+
+    def evaluate_circuit_time(self, hw, engine, builder):
+        qatfile = InstructionEmitter().emit(builder.instructions, hw)
+        res = engine.create_duration_timeline(qatfile.instructions)
+        # pulse channels could have different block times
+        return max([val[-1].end * pc.block_time for pc, val in res.items()])
```
hamidelmaazouz, line 805: 👍🏽  but make sure you distinguish samples from time in the PositionData specification. They are most of the time confused as the block size is 1 by default.

tests/qat/test_quantum_backend.py
```diff
@@ -714,6 +714,163 @@ def test_create_duration_timeline_mapping(self):
...
+        circs = [
+            get_builder(hw).X(q1, np.pi / 2.0),
+            get_builder(hw).X(q2, np.pi),
+            get_builder(hw).cnot(q1, q2),
+            get_builder(hw).measure(q1),
+            get_builder(hw).measure(q2),
+        ]
+
+        ts = [self.evaluate_circuit_time(hw, engine, circ) for circ in circs]
+        assert np.isclose(ts[0] + ts[1] + ts[2] + max(ts[3], ts[4]), maxtime)
```
hamidelmaazouz, line 872: can you explain why `max(ts[3], ts[4])` ?

tests/qat/test_quantum_backend.py
```diff
@@ -714,6 +714,163 @@ def test_create_duration_timeline_mapping(self):
...
+
+        ts = [self.evaluate_circuit_time(hw, engine, circ) for circ in circs]
+        assert np.isclose(max(ts[0], ts[1]) + ts[2] + max(ts[3], ts[4]), maxtime)
+
+    @pytest.mark.parametrize(
+        "hw", [get_test_model(), get_default_echo_hardware(), get_default_RTCS_hardware()]
+    )
+    def test_duration_timeline_compare_sync(self, hw):
+        # Tests that the duration of individual circuit elements matches that
+        # of the full circuit when syncs are used.
```
hamidelmaazouz, line 845: 👍🏽

hamidelmaazouz (approved):
thanks !

Regarding visualisation, not sure if `examples/plot_pulse_buffer.py` is still relevant, but could you run it to make sure it still runs as expected ?

tests/qat/test_quantum_backend.py
lcauser-oqc: To clarify, the test currently looks for all instructions that are _not_ Delays, and checks that they start and end at the same time as the instruction set without the sync. We ignore delays because the addition sync adds additional delay instruction to measure / acquire targets (which in practice can be squashed to a single delay for each target, but the result it the same).

However, we can remove `inst-start != inst.end` (I can't remember my reasoning for adding this in first place...).
Also added the check on acquires as you suggest, thanks!

tests/qat/test_quantum_backend.py
lcauser-oqc: Sorry, my comment for the test made it confusing. What I meant was: the test checks that instructions don't overlap or leave gaps in time (so the end of one instruction should be the beginning of the next).

But I agree, having a test that that compares targets where instructions should align will be valuable. I'll add that :smile:

tests/qat/test_quantum_backend.py
lcauser-oqc: Very good point, addition test added!

PR #185, opened 2024-09-18
Dynamically handle sync reset lifecycle

hamidelmaazouz:


src/qat/purr/backends/qblox/device.py
```diff
@@ -253,6 +253,11 @@ def start_playback(self, repetitions: int, repetition_time: float):
             raise ValueError("No resources allocated. Install packages first")
 
         results = {}
+
+        for module, allocations in self._resources.items():
```
keriksson-rosenqvist, line 257: Is it strictly necessary to redo this iteration over and over? :thinking:

src/qat/purr/backends/qblox/device.py
```diff
@@ -285,7 +290,11 @@ def start_playback(self, repetitions: int, repetition_time: float):
                         ) / sequencer.integration_length_acq()
 
                     sequencer.delete_acquisition_data(all=True)
-                    sequencer.sync_en(False)
```
keriksson-rosenqvist, line 288: Here you're replacing a previous call with one in a separate iteration but there was no previous `sequencer.sync_en(True)` call. Could you explain what this does?

keriksson-rosenqvist (approved):
Some open questions but generally LGTM.

src/qat/purr/backends/qblox/device.py
bgsach: This would look cleaner pulled out as a short helper function with a good name...

It looks like the code flow is something like ...

```
self._sync_resources(True)
*do stuff*
self._sync_resources(False)
```

bgsach (changes requested):

src/qat/purr/backends/qblox/device.py
hamidelmaazouz: previously sync was being set as part of the configuration step. Not any more, everything must remain false, except for the involved module sequencer in the experiment

src/qat/purr/backends/qblox/device.py
hamidelmaazouz: yep, it's necessary ^^

hamidelmaazouz: > This would look cleaner pulled out as a short helper function with a good name...
> 
> It looks like the code flow is something like ...
> 
> ```
> self._sync_resources(True)
> *do stuff*
> self._sync_resources(False)
> ```

it's a nice but minor improvement given the scale of other things, it'll part of another PR

bgsach (approved):

PR #186, opened 2024-09-18
randomized_benchmarking bug

lcauser-oqc:
`randomized_benchmarking` throws error if lengths was given as key argument.


src/qat/purr/utils/benchmarking.py
```diff
@@ -21,6 +21,8 @@ def randomized_benchmarking(hardware, nseeds, *args, **kwargs):
     if "lengths" not in kwargs:
         # TODO: Find out why length 10 generates some broken DAG when re-parsed
         lengths = [1, 2, 4]
+    else:
+        lengths = kwargs.pop("lengths")
```
bgsach, line 25: wouldn't it be cleaner to do

```lengths = kwargs.get('lengths', [1,2,4]) ```

bgsach (approved):
LGTM

PR #187, opened 2024-09-18
Changing CODEOWNERS file

chemix-lunacy:
While I may occasionally put up PRs around integration with my other projects it's rare I'm involved in QAT directly nowadays.


bgsach (approved):
Thank you for all your hard work on QAT!

PR #189, opened 2024-09-18
Only make releases from tags on main where poetry version matches tag

bgsach:
This changes our package publishing action to run on a tag push. Previously it was intended to run on a 'tag push on main' but in fact it was triggering on a tag OR main. The new flow will trigger on any tag but check that tag is a commit on main and abort if not (thanks to @matt-oqc for this fix on QAT-RPC).

It also checks that the poetry version (in pyproject.toml) precisely matches the tag.

Tag extraction uses https://stackoverflow.com/a/72959712/8179249

### DIFF
--- a/.github/workflows/package-publish.yml
+++ b/.github/workflows/package-publish.yml
@@ -4,7 +4,6 @@ on:
   push:
     tags: [ '[0-9]+.[0-9]+.[0-9]+' ]
 jobs:
-  # @see https://stackoverflow.com/a/72959712/8179249
   check-current-branch:
     runs-on: ubuntu-latest
     outputs:
--- a/.github/workflows/package-publish.yml
+++ b/.github/workflows/package-publish.yml
@@ -2,14 +2,35 @@ name: Publish package to PyPI and TestPyPI
 
 on:
   push:
-    branches: [ main ]
     tags: [ '[0-9]+.[0-9]+.[0-9]+' ]
-
 jobs:
+  # @see https://stackoverflow.com/a/72959712/8179249
+  check-current-branch:
+    runs-on: ubuntu-latest
+    outputs:
+      branch: ${{ steps.check_step.outputs.branch }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Get current branch
+        id: check_step
+        # 1. Get the list of branches ref where this tag exists
+        # 2. Remove 'origin/' from that result
+        # 3. Put that string in output
+        # => We can now use function 'contains(list, item)''
+        run: |
+          raw=$(git branch -r --contains ${{ github.ref }})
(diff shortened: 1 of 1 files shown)


daria-oqc (approved):

lcauser-oqc (approved):

keriksson-rosenqvist (approved):
LGTM
We could skip the need for the tag to be on main if we needed to, or have a `branch == main | tag-match` if we wanted that additional check.

keriksson-rosenqvist:
If these actions are already in `qat-rpc` we should centralise them. But not blocking the merge as long as that gets done at some point.

matt-oqc:
> If these actions are already in `qat-rpc` we should centralise them. But not blocking the merge as long as that gets done at some point.

After some research on this matter, public repos cannot make use of reusable workflows within private repos, and there [is a note on the `pypa/gh-action-pypi-publish` action](https://github.com/pypa/gh-action-pypi-publish?tab=readme-ov-file#trusted-publishing) that trusted publishing cannot be used as part of a reusable workflow.

We should centralise as much as we can though, and preparatory phases of this publishing workflow can certainly be maintained here under e.g. a `oqc-community/actions` repo.

.github/workflows/package-publish.yml
```diff
@@ -2,14 +2,35 @@ name: Publish package to PyPI and TestPyPI
 
 on:
   push:
-    branches: [ main ]
     tags: [ '[0-9]+.[0-9]+.[0-9]+' ]
-
 jobs:
+  # @see https://stackoverflow.com/a/72959712/8179249
```
hamidelmaazouz, line 7: minor: I'm reluctant to leave links and commented code in the source tree. I would add it to the description of the associated ticket (if any) (unless it's a temporary change)

hamidelmaazouz (approved):
tests ? 😛

.github/workflows/package-publish.yml
bgsach: I concur, will resolve

bgsach:
> tests ? 😛

If you have a suggestion for how to do this easily, I'd be up for it. Debugging pipelines is a pain (but there is a linter available as a vs code extension which really helps).

matt-oqc:
> > tests ? 😛
> 
> If you have a suggestion for how to do this easily, I'd be up for it. Debugging pipelines is a pain (but there is a linter available as a vs code extension which really helps).

https://github.com/rhysd/actionlint is a tool I found

bgsach:
> > > tests ? 😛
> > 
> > 
> > If you have a suggestion for how to do this easily, I'd be up for it. Debugging pipelines is a pain (but there is a linter available as a vs code extension which really helps).
> 
> https://github.com/rhysd/actionlint is a tool I found

I use this https://marketplace.visualstudio.com/items?itemName=GitHub.vscode-github-actions

PR #191, opened 2024-09-19
Minor pyproject fixes.

keriksson-rosenqvist:


bgsach (approved):
You even fixed a mistake I made in my own name! (Clearly subconsciously I don't think of myself as important enough to have a capital)

daria-oqc (approved):
Thanks!

PR #192, opened 2024-09-20
COMPILER-113 and COMPILER-115: Add rigidity to QatConfig settings

daria-oqc:


src/qat/purr/qatconfig.py
```diff
@@ -10,5 +11,13 @@ class QatConfig(BaseSettings):
     MAX_REPEATS_LIMIT: int = 100_000
     """Max number of repeats / shots to be performed in a single job."""
 
+    @field_validator("MAX_REPEATS_LIMIT", mode="before")
+    def check_max_repeats_limit(cls, value):
+        if value and not isinstance(value, int):
+            raise TypeError(
+                f"MAX_REPEATS_LIMIT must be an int, got {type(value).__name__}."
+            )
+        return value
```
bgsach: you get this 'for free' in pydantic, all type hints are enforced.

tests/qat/test_qatconfig.py
```diff
@@ -2,11 +2,32 @@
 # Copyright (c) 2023 Oxford Quantum Circuits Ltd
 import pytest
 
-from qat import qatconfig
 
+# from qat import qatconfig
+@pytest.fixture
+def qatconfig():
+    """We import qatconfig as a fixture so we can test mocking the environment."""
+    from qat import qatconfig
```
bgsach, line 11: I think we realised we don't need this fixture.

tests/qat/test_qatconfig.py
```diff
@@ -2,11 +2,32 @@
...
+    # Test direct change of repeats limit.
+    qatconfig.MAX_REPEATS_LIMIT = repeats_limit
+    assert qatconfig.MAX_REPEATS_LIMIT == repeats_limit
+
+
+@pytest.mark.parametrize("repeats_limit", [10_000, 50_000, 16_874, 100_000])
+def test_change_env_max_repeats_limit(qatconfig, monkeypatch, repeats_limit):
+    monkeypatch.setenv("QAT_MAX_REPEATS_LIMIT", str(repeats_limit))
+    # Default value cannot change if env variable changes after instantiation.
+    assert qatconfig.MAX_REPEATS_LIMIT == 100_000
```
bgsach, line 34: probably best to explicitly set  qatconfig.MAX_REPEATS_LIMIT at the start of the test because the default might change.

bgsach (changes requested):

daria-oqc:
@bgsach I have incorporated your comments. The removal of `field_validator` was already pushed, but probably was not refreshed during your review.

Additionally - as I do not see any reason why `MAX_REPEATS_LIMIT` cannot be None, I have added this functionality + tests accordingly.

tests/qat/test_qatconfig.py
```diff
@@ -1,12 +1,30 @@
...
+    # Test direct change of repeats limit.
+    qatconfig.MAX_REPEATS_LIMIT = repeats_limit
+    assert qatconfig.MAX_REPEATS_LIMIT == repeats_limit
+
+
+@pytest.mark.parametrize("repeats_limit", [None, 10_000, 16_874, 50_000, 100_000])
+def test_change_env_max_repeats_limit(monkeypatch, repeats_limit):
+    monkeypatch.setenv("QAT_MAX_REPEATS_LIMIT", str(repeats_limit))
+    # Default value cannot change if env variable changes after instantiation.
+    assert qatconfig.MAX_REPEATS_LIMIT == MAX_REPEATS_LIMIT
```
bgsach, line 30: I think we need a test that setting via the env does work "at the start"...

```
from qat.purr.qatconfig import QatConfig
def test_set_via_env(monkeypatch):
    NEW_LIMIT = 15321
    monkeypatch.setenv("QAT_MAX_REPEATS_LIMIT", str(NEW_LIMIT))
    newconfig = QatConfig()
    assert newconfig.MAX_REPEATS_LIMIT == NEW_LIMIT
   ```
   
   something like that?

bgsach (changes requested):

tests/qat/test_qatconfig.py
daria-oqc: Agreed, and additionally check if the existing `qatconfig` is not changed.

This, together with enforced values for `MAX_REPEATS_LIMIT`, is implemented in the latest commit.

bgsach (approved):
LGTM

PR #193, opened 2024-09-23
Caching full ids

lcauser-oqc:
Related to #178 

**Changes:**
- purr/compiler/devices.py: caching `id` and `full_id`, deleting the cached ids
- purr/compiler/hardware_models.py: `delete_cache()` to clean the cached ids for all components,
- Tests to ensure caching works as intended...

**A few things to consider (please give your thoughts):**
- The cached `full_id` is still retrieved using a function call: this shouldn't break any dependencies, but obviously has some overhead. Would be happy to change.
- I'm of the opinion that the architecture of hardware models should be immutable (including ids) but have calibratables that can be changed and having the functionality to change IDs like I've added here shouldn't be encouraged... However, right now, these do serve as a useful way to test that the clearing the cache does work as intended. I have added them just to the tests.

Some data on the kind of speeds up we see (why caching the ids is important): N qubits each with 100 cnot layers and 100 U layers.
|  | N=10 before | N=10 after | N=2 before | N=2 after |
| --- | --- | --- | --- | --- |
| Building the circuit | 0.434s | 0.312s | 0.0476s | 0.0396s |
| Optimising the circuit | 0.104s | 0.0352s | 0.0149s | 0.00627s |
| Creating the timeline | 0.332s | 0.156s | 0.0589s | 0.0280s |


src/qat/purr/compiler/devices.py
```diff
@@ -5,6 +5,7 @@
 import os
 import sys
 from enum import Enum, auto
+from functools import cached_property
```
daria-oqc, line 8: If we are moving towards `pydantic` functionality in qat, would it be a good idea to start using `QuantumComponent` as a child class of `BaseModel` and use their functionality for cached properties? See [pydantic docu](https://docs.pydantic.dev/2.0/usage/computed_fields/)

hamidelmaazouz (changes requested):
Love that you're attempted to manage the cache yourself, sounds great but it overfits the complexity the whole problem deserves. I also don't want to introduce`cached_full_id` as a new member or inherit the headaches of managing their changes, all of that plays against simplicity and ease to debug if anything ever goes wrong.


Instead, have a look at:
- Last four sections of `@cached_property` https://docs.python.org/3/library/functools.html#functools.cached_property
- `@lru_cache` and more useful info at https://docs.python.org/3/faq/programming.html#how-do-i-cache-method-calls
- Seemingly unrelated, but checkout the story about thread safety of `@cached_property`at https://discuss.python.org/t/finding-a-path-forward-for-functools-cached-property/23757/17?page=2.

hamidelmaazouz:
Here's an example:

```
class SomeClass:
    def __init__(self, number):
        self._number = number

    def __eq__(self, other):
        return self._number == other._numer

    def __hash__(self):
        return hash(self._number)

    @property
    @lru_cache
    def doubled(self):
        return self.number * 2

    @property
    def number(self):
        return self._number

    @number.setter
    def number(self, val):
        self._number = val


obj = SomeClass(1)
assert obj.doubled == 2 * 1
assert obj.doubled == 2 * obj.number

obj.number = 5 # Access through a setter : Okay
assert obj.doubled == 2 * 5
assert obj.doubled == 2 * obj.number

obj._number = 2 # Access through a private field : Okay
assert obj.doubled == 2 * obj.number
assert obj.doubled == 2 * 2

```

src/qat/purr/compiler/devices.py
bgsach: I'd love to do this but I think we should probably prototype out a switch to pydantic first.

lcauser-oqc:
> Love that you've attempted to manage the cache yourself, sounds great but it overfits the complexity the whole problem deserves. I also don't want to introduce`cached_full_id` as a new member or inherit the headaches of managing either's cached or non cached property changes, all of that plays against simplicity and ease to debug if anything ever goes wrong.
> 
> Instead, have a look at:
> 
> * Last four sections of `@cached_property` https://docs.python.org/3/library/functools.html#functools.cached_property
> * `@lru_cache` and more useful info at https://docs.python.org/3/faq/programming.html#how-do-i-cache-method-calls
> * Seemingly unrelated, but checkout the story about thread safety of `@cached_property`at https://discuss.python.org/t/finding-a-path-forward-for-functools-cached-property/23757/17?page=2.

Thanks for the nice suggestions!

After some playing with it, I think there will be some problems with using `@lru_cache` that would require a solution as specific as the one we have here. The main difficulty is that caching the outcome of `full_id` will rely on on `__hash__`. In your example, `__hash__` only depends on `number` so it would make sense. Here, we need it to depend on `full_id`, which will get us stuck in a recursion loop. If we try to hash using something other than the `full_id`, e.g. the `partial_id`, then we risk not having full coverage to any possible changes to the `full_id` (in the case that we hash with `partial_id`, the cached `full_id` would not be updated if the `id` of the physical channel changes).

hamidelmaazouz:
> Here, we need it to depend on full_id, which will get us stuck in a recursion loop. If we try to hash using something other than the full_id, e.g. the partial_id, then we risk not having full coverage to any possible changes to the full_id (in the case that we hash with partial_id, the cached full_id would not be updated if the id of the physical channel changes).

Good point on the recursion idea, have you tried it ?
Another simple solution is keep the @cached_property decorator, but enforce immutability all components composing the full_id

hamidelmaazouz:
I am more in favour of enforcing immutability, but allow users to copy-construct pulse channels if they need to

lcauser-oqc:
I did try using the `lru_cache`: changing the hash to the `partial_id` instead of the `full_id` did break the loop, but as anticipated, did not allow for the `full_id` to be updated if the id of a physical channel is changed. 

I am also in favour of enforcing immutability. However, there is need to dynamically change the hardware model (@bgsach might have details on this).

PR #194, opened 2024-09-23
COMPILER-103 Add compiler-config as a qat dependency.

daria-oqc:
- Fix imports to refer to `compiler-config` package.
- Remove tests for serialiser that belong to `compiler-config`. See https://github.com/oqc-community/compiler-config/pull/2 .
- Remove minimal package containing `config.py` and `serializer.py` since this functionality is now provided by `compiler-config`.


pyproject.toml
```diff
@@ -45,6 +45,7 @@ jsonpickle = ">=2.2.0"
 qblox-instruments = "^0.14.0"
 lark-parser = "^0.12.0"
 pydantic-settings = ">=2.5.2"
+compiler-config = { git = "https://github.com/oqc-community/compiler-config.git", branch = "main" }
```
daria-oqc, line 48: This is a temporary solution until we have a pypi release of compiler-config with a specific version number.

bgsach: I'm happy with this in the interim

PR #195, opened 2024-09-24
Bugfix: Overriding open pulse intrinsics  

lcauser-oqc:
Changes:
- Moved the ecr and cnot further within the if-block.
- Added tests to check that they can be overridden. 
- Added a test to ensure that adding unitaries execute.

Note: `U` isn't supported in openpulse grammar, so I left this alone.


tests/qat/files/qasm/qasm3/u_test.qasm
hamidelmaazouz, line 1: super minor: rename file to `u_gate`

tests/qat/files/qasm/qasm3/u_test.qasm
```diff
@@ -0,0 +1,6 @@
+OPENQASM 3;
+defcalgrammar "openpulse";
+
+
+u(pi, pi/2, pi/4) $0;
+u(-pi/2, pi, -pi/4) $1;
```
hamidelmaazouz, line 6: can you add more non common angles or edge cases ?

tests/qat/test_qasm.py
```diff
@@ -228,6 +228,46 @@ def test_cx_override(self):
...
+                    inst
+                    for inst in result.instructions
+                    if hasattr(inst, "shape")
+                    and (inst.shape is PulseShapeType.EXTRA_SOFT_SQUARE)
+                ]
+            )
+            == 2
+        )
+
+    def test_u(self):
```
hamidelmaazouz, line 265: `test_u_gate`

tests/qat/test_qasm.py
```diff
@@ -228,6 +228,46 @@ def test_cx_override(self):
...
+        hw = get_default_echo_hardware(2)
+        parser = Qasm3Parser()
+        result = parser.parse(get_builder(hw), get_qasm3("u_test.qasm"))
+        assert len(result.instructions) > 0
```
hamidelmaazouz, line 269: add more assertions that actually verifies the guts of the result

tests/qat/test_qasm.py
```diff
@@ -228,6 +228,46 @@ def test_cx_override(self):
...
+            len(
+                [
+                    inst
+                    for inst in result.instructions
+                    if hasattr(inst, "shape")
+                    and (inst.shape is PulseShapeType.EXTRA_SOFT_SQUARE)
+                ]
+            )
+            == 2
+        )
```
hamidelmaazouz, line 263: same here, more finer-detail assertions

tests/qat/test_qasm.py
```diff
@@ -228,6 +228,46 @@ def test_cx_override(self):
...
```
hamidelmaazouz, line 246: same here

tests/qat/test_qasm.py
```diff
@@ -228,6 +228,46 @@ def test_cx_override(self):
...
+        # tests that there are 2 extra_soft_square pulses from the custom def
+        assert (
+            len(
+                [
+                    inst
+                    for inst in result.instructions
+                    if hasattr(inst, "shape")
+                    and (inst.shape is PulseShapeType.EXTRA_SOFT_SQUARE)
+                ]
+            )
```
hamidelmaazouz, line 244: not robust enough, test should test parsing result is it matches what's in the qasm, and then we can add assertions on the type of the generated instructions for example, also test the targets of the instruction: expect a Pulse to target logical pulse channel (drive for example) associated to the qubits in parameters ... lots of things you can check

hamidelmaazouz (commented):
A few comments on tests, but looking good !

lcauser-oqc:
- Tests for U gates: added a wider range of parameters including special cases, compared them to the same circuit made using qat to ensure they match.
- Tests for cnot/ecr overridding: checks that that the overridden gate targets channels for each qubit, and that the instructions for the non overridden gates match the expected instructions (again generated using qat)

tests/qat/test_qasm.py
bgsach: I'm not sure what the the check from line 254 onwards adds doesn't it essentially check that hw.create_builder() creates a fresh builder which doesn't "hang on" to the overrides? 

Also the override tests (test_ecr_override, test_cx_override, test_cx_override) look like they could be parameterised by the gate? `get_qasm3(f"{gate}_override_test.qasm")` or similar (we could actually parameterise the .qasm files too with jinja2 etc. but I think that's for another day).

hamidelmaazouz (approved):
Thanks !

tests/qat/test_qasm.py
lcauser-oqc: The tests only override the gate for qubits (1, 2), so we want to make sure that gate on qubits (0, 1) and (2, 3) are unchanged, i.e., equivalent to what we will expect to see from the circuit builder. So we check that the instruction lists match.

Nice suggestion, on it now! :)

bgsach (approved):
LGTM

PR #198, opened 2024-09-27
COMPILER-118 Add validation for qatconfig 'system' settings.

daria-oqc:


src/qat/purr/qatconfig.py
```diff
@@ -86,10 +87,16 @@ def check_disable_pulse_duration_limits(cls, DISABLE_PULSE_DURATION_LIMITS):
             )
         return DISABLE_PULSE_DURATION_LIMITS
 
+    def validate(self, compiler_config: CompilerConfig):
```
bgsach, line 90: I think this should be called validate_compiler_config just to make it clear what it's validating?

src/qat/purr/compiler/runtime.py
```diff
@@ -383,6 +384,7 @@ def execute_instructions(
 ):
     if config is None:
         config = CompilerConfig()
+    qatconfig.validate(config)
     config.validate(hardware)
```
hamidelmaazouz, line 388: complexity of validation is building up here, lots of things that look similar but do unclear operations

goes in line with a comment here: https://github.com/oqc-community/qat/pull/271#discussion_r1859257014

bgsach: I agree with the sentiment but I think we need to get the feature to main and then iteratively refactor. Let's land some of the big ticket items (pipelines, hardware_model refactors) then refactor this later.

bgsach:
Closed in favour of internal PR

PR #199, opened 2024-09-27
[EXPERIMENTAL] Prototype Pydantic HardwareModels

daria-oqc:


hamidelmaazouz (changes requested):
Since you're already pioneering QAT refactoring (staring with the HW models), I don't think it's a good idea to replay the same legacy module structure or the same class hierarchy. Let's make it a totally brand new rewrite. Here we must distinguish "compilation" from "execution" and clearly understand what "runtime" means and does.

My suggestion is to build a whole new structure outside of the `qat.purr` package:
- `qat.purr` is legacy, let it burn and die a painful death
- `qat.ir.model` for the hardware model: just the quantum side, try to strip it out of any hints to the control stack.
- `qat.ir.instructions` for qat instructions or a better equivalent
- `qat.runtime` for any data structure, algorithm, or operation supporting execution of the builder, post processing results and formatting them.
- `qat.frontend` to host the usual stuff: parsing, walking, transformation, analysis, and validation. It's actually not the same as the current frontend
- `qat.backend` or `qat.target` to contain all backend related problems such as codegen surrounded by all lowering passes
- `qat.driver` for any execution / interpretation. I classical compilers, "driver" is usually reserved for tools facing the the customer such as `gcc` or `clang`. This is roughly the same as the current `frontend`

daria-oqc:
Closed in favour of #240.

PR #201, opened 2024-10-01
Restore legacy compiler config compatibility.

keriksson-rosenqvist:
Aliasing old compiler config path and raise `DeprecationWarning`s when old path is used.
Also restore legacy serialised versions and add v2.


src/qat/purr/compiler/config.py
```diff
@@ -0,0 +1,12 @@
+# SPDX-License-Identifier: BSD-3-Clause
+# Copyright (c) 2023 Oxford Quantum Circuits Ltd
+import warnings
+
+warnings.simplefilter("default", DeprecationWarning)
```
daria-oqc, line 5: 👍🏻

daria-oqc (approved):

lcauser-oqc (approved):
Lgtm!

PR #202, opened 2024-10-01
Performance improvement: caching the full_id in PulseChannels

lcauser-oqc:
Hashing a `PulseChannel` makes expensive calls to construct a `full_id`, leading to serious performance decreases. To circumvent this, we:
- Save the constructed `full_id` as a private member of the `PulseChannel`.
- Use setters for `id` and `PhysicalChannel` to update the `full_id` (these are the instances where a `full_id` changes).
- `hash(PulseChannelView) = hash(PulseChannelView.pulse_channel)`.
- Changes to eliminate places where we unnecessarily call `full_id`.


src/qat/purr/compiler/hardware_models.py
```diff
@@ -305,8 +310,11 @@ def get_devices_from_pulse_channel(self, id_: str):
     def get_pulse_channel_from_id(self, id_: str):
         return self.pulse_channels.get(id_)
 
-    def get_devices_from_physical_channel(self, id_: str):
-        physical_channel = self.get_physical_channel(id_)
+    def get_devices_from_physical_channel(
+        self, physical_channel: Union[str, PhysicalChannel]
+    ):
+        if isinstance(physical_channel, str):
+            physical_channel = self.get_physical_channel(physical_channel)
```
hamidelmaazouz, line 317: the `self.get_physical_channel()` method is too short, please inline it here, no need for extra call:

```
    def get_physical_channel(self, id_: str):
        return self.physical_channels.get(id_, None)
```

src/qat/purr/compiler/hardware_models.py
```diff
@@ -370,7 +378,7 @@ def _resolve_qb_pulse_channel(self, chanbit: Union[Qubit, PulseChannel]):
         if isinstance(chanbit, Qubit):
             return chanbit, chanbit.get_default_pulse_channel()
         else:
-            quantum_devices = self.get_devices_from_pulse_channel(chanbit.full_id())
+            quantum_devices = self.get_devices_from_pulse_channel(chanbit)
```
hamidelmaazouz, line 381: also adjust the if block slightly downwards

```
            if len(primary_devices) > 1:
                raise NotImplementedError(
                    f"Too many devices use the channel with id {chanbit.full_id()} as "
                    "a default channel to resolve a primary device."
                )
```

hamidelmaazouz (commented):
Good approach if you're planning to ditch `full_id` calls altogether. Otherwise you'd still need to keep an association of the (private (and hopefully immutable)) uuid to the full_id() which is based on mutable fields

bgsach (approved):
LGTM subject to resolution of Hamid's outstanding comments

keriksson-rosenqvist (approved):
LGTM

src/qat/purr/compiler/devices.py
```diff
@@ -404,6 +405,29 @@ def __init__(
...
             )
 
+    @property
+    def id(self):
+        return self._id
+
+    @id.setter
+    def id(self, val):
+        self._id = val
+        if hasattr(self, "_physical_channel"):
```
hamidelmaazouz, line 415: is this for the PulseChannelView problem ?

hamidelmaazouz (approved):
Please squash all commit into one ?

src/qat/purr/compiler/devices.py
lcauser-oqc: Thanks for checking it over! :smile: 

It's because constructing the full id needs both `_id` and `_physical_channel` to be set: in the constructor, `id` is called before the physical channel is set.

PR #203, opened 2024-10-02
Wh/remove channels fix

keriksson-rosenqvist:
Removes channels that just Wait and phase Reset since these are doing nothing!

### DIFF
--- a/src/qat/purr/compiler/execution.py
+++ b/src/qat/purr/compiler/execution.py
@@ -472,11 +472,16 @@ class QuantumExecutionEngine(InstructionExecutionEngine):
                         current_duration + position_data.instruction.duration
                     )
 
-        # Strip timelines that only hold delays, since that just means nothing is
+        # Strip timelines that only hold delays and phase resets, since that just means nothing is
         # happening on this channel.
-        for key, timeline in dict(results).items():
-            if all(isinstance(pos_data.instruction, Delay) for pos_data in timeline):
-                del results[key]
+        results = {
+            key: timeline
+            for key, timeline in dict(results).items()
+            if not all(
+                isinstance(pos_data.instruction, (Delay, PhaseReset, PhaseShift))
+                for pos_data in timeline
+            )
+        }
 
         if final_positions := [
             final_position[-1].end for final_position in list(results.values())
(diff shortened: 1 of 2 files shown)


tests/qat/test_quantum_backend.py
```diff
@@ -632,3 +638,28 @@ def test_execution_batching(self, repeat_count, repeat_limit, expected_batches):
...
         generated_batches = engine._generate_repeat_batches(repeat_count)
         assert generated_batches == expected_batches
+
+    def test_empty_channels_removed(self):
+        hw = get_test_model()
+        qubit = hw.get_qubit(0)
+
+        builder = (
+            get_builder(hw).X(qubit, np.pi / 2.0).measure_mean_z(qubit).synchronize(qubit)
+        )
```
hamidelmaazouz, line 648: not enough testing, add more builders that involve for example a variety of pulse channels: drive, measure, acquire, second state, ...

src/qat/purr/compiler/execution.py
```diff
@@ -468,7 +468,10 @@ def create_duration_timeline(self, instructions: List[QuantumInstruction]):
         # Strip timelines that only hold delays, since that just means nothing is
         # happening on this channel.
         for key, timeline in dict(results).items():
```
hamidelmaazouz, line 470: minor: instead of a new dict:

```
dict = {k, v for k, v in dict.items() if somthing_about(k, v)}
```

hamidelmaazouz (commented):
Happy with the solution, just needs more testing

tests/qat/test_quantum_backend.py
keriksson-rosenqvist: @willh727-OQC

willh727-OQC: They are all added when you add a qubit! Do I need more tests for just check these channels?

src/qat/purr/compiler/execution.py
willh727-OQC: Done, with the formatting it does take up more lines tho

hamidelmaazouz (approved):

PR #205, opened 2024-10-03
Add override option to pulse duration limit checks.

daria-oqc:


src/qat/purr/compiler/execution.py
```diff
@@ -244,10 +245,15 @@ def validate(self, instructions: List[Instruction]):
             if isinstance(inst, (Pulse, CustomPulse)):
                 duration = inst.duration
                 if isinstance(duration, Number) and duration > MaxPulseLength:
-                    raise ValueError(
-                        f"Max Waveform width is {MaxPulseLength} s "
-                        f"given: {inst.duration} s"
-                    )
+                    if (
+                        not qatconfig.DISABLE_PULSE_DURATION_LIMITS
```
keriksson-rosenqvist, line 249: Looks good. Could we also add it in the variable case, see row 268 below.

keriksson-rosenqvist (changes requested):

src/qat/purr/compiler/execution.py
daria-oqc: Done, covered with a test as well. In the meantime, I refactored some qatconfig tests to make sure that the qatconfig object gets reset to its default state after the qatconfig tests.

keriksson-rosenqvist (approved):

Issue #129, opened 2024-07-10
Intermittent OOM error when running QiskitEngine


hamidelmaazouz:
**Describe the bug**
```
ERROR:  [Experiment 0] Insufficient memory to run circuit circuit-313 using the statevector simulator.
Required memory: 524288M, max memory: 15981M
```

**Describe what you expected to happen**
This failure is intermittent and happens when running the Qiskit-based QASM simulator unit tests in CI.

**Add supporting information**
Intermittently observable in CI unit tests. The failure is closely related to the simulation method. Some methods are resource hungry.

**Actions**
* Deep dive and report about Qiskit's `AerSimulator` simulation methods
* Identify resource-hungry methods
* One workaround is simply add more memory to the GH runner. Although it's probably worthwhile to make simulation methods configurable in QAT so that it's easy to avoid expensive ones.

hamidelmaazouz:
This bug is being addressed here: https://github.com/oqc-community/qat/pull/206. Cheers @lcauser-oqc

PR #206, opened 2024-10-03
Extending qiskit simulator capabilities

lcauser-oqc:
Changes made:
- Options for the `AerSimulator` backend can now be passed via qatconfig.
- A fall-back sequence can be provided for `AerSimulator` if the given method fails.
- If simulations using `automatic` or `statevector` fail due to a lack of resources, by default `AerSimulator` will fall back on its MPS backend with a controlled bond dimension. See #129.
- Some basic documentation for the Qiskit backend.

### DIFF
--- a/src/qat/purr/backends/qiskit_simulator.py
+++ b/src/qat/purr/backends/qiskit_simulator.py
@@ -344,7 +344,7 @@ class QiskitEngine(InstructionExecutionEngine):
                 task_results[creg] = creg_result
         else:
             task_results = trimmed
-        
+
         return (
             (task_results, results.results[0].metadata)
             if qiskitconfig.ENABLE_METADATA
--- a/src/qat/purr/backends/qiskit_simulator.py
+++ b/src/qat/purr/backends/qiskit_simulator.py
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: BSD-3-Clause
 # Copyright (c) 2023 Oxford Quantum Circuits Ltd
 import re
-from typing import Any, Dict, List, Optional, Tuple, Union
+from typing import List, Optional, Tuple, Union
 
 import numpy as np
 from compiler_config.config import ErrorMitigationConfig, ResultsFormatting
@@ -11,6 +11,7 @@ from qiskit.transpiler.passes import CheckMap
 from qiskit_aer import AerSimulator
 from qiskit_aer.backends.backendconfiguration import AerBackendConfiguration
 
+from qat import qatconfig
 from qat.purr.backends.echo import (
     Connectivity,
     add_direction_couplings_to_hardware,
@@ -33,6 +34,7 @@ from qat.purr.compiler.runtime import QuantumRuntime
 from qat.purr.utils.logger import get_default_logger
 
 log = get_default_logger()
+qiskitconfig = qatconfig.SIMULATION.QISKIT
 
 
 def get_default_qiskit_hardware(
@@ -41,6 +43,12 @@ def get_default_qiskit_hardware(
     strict_placement=True,
     connectivity: Optional[Union[Connectivity, List[Tuple[int, int]]]] = None,
 ) -> "QiskitHardwareModel":
+    """
+    Creates a hardware model compatible with the Qiskit simulator.
+
+    If `strict_placement=True`, circuits can only be executed when circuit
+    intructions act on adjacent qubits in the coupling map.
+    """
     model = QiskitHardwareModel(qubit_count, noise_model)
     model.strict_placement = strict_placement
 
@@ -229,7 +237,27 @@ class QiskitEngine(InstructionExecutionEngine):
     def run_calibrations(self, qubits_to_calibrate=None):
         pass
 
-    def execute(self, builder: QiskitBuilder) -> Dict[str, Any]:
+    def _run_simulator(self, builder, aer_config, method, coupling_map):
+        qasm_sim = AerSimulator(
+            aer_config,
+            noise_model=builder.model.noise_model,
+            method=method,
+            **qiskitconfig.OPTIONS,
+        )
+        job = qasm_sim.run(
+            transpile(builder.circuit, qasm_sim, coupling_map=coupling_map),
+            shots=builder.shot_count,
+        )
+        return job.result()
+
+    def execute(self, builder: QiskitBuilder):
+        """
+        Execute a circuit using Qiskit's AerSimulator as a backend. Options for the
+        simulator can be provided using qatconfig.SIMULATION.QISKIT.
+
+        For more information, see:
+        https://docs.quantum.ibm.com/api/qiskit/0.37/qiskit.providers.aer.AerSimulator
+        """
         if not isinstance(builder, QiskitBuilder):
             raise ValueError(
                 "Contravariance is not possible with QiskitEngine. "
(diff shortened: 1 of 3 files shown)


src/qat/purr/qatconfig.py
```diff
@@ -50,6 +50,20 @@ def check_disable_pulse_duration_limits(cls, DISABLE_PULSE_DURATION_LIMITS):
...
+    The default settings for using the MPS backend in the Qiskit Simulator.
+    """
+
+    model_config = SettingsConfigDict(env_previx="QAT_MPS_")
+    MAX_BOND_DIMENSION: int = 128
+    TRUNCATION: float = 1e-12
+
+
+qatmpsconfig = QatMPSConfig()
+
```
lcauser-oqc, line 65: This might be misplaced in the qatconfig file. More than happy to rehome it to qiskit_simulator

daria-oqc: I think it might be better to keep QatConfig for system-wide settings, indeed, and rehome this one to qiskit_simulator.

bgsach (changes requested):
Looks good except for minor comments

src/qat/purr/backends/qiskit_simulator.py
```diff
@@ -29,12 +30,33 @@
...
+
+    model_config = SettingsConfigDict(env_previx="QAT_MPS_")
+    MAX_BOND_DIMENSION: int = 128
+    """Default maximum bond dimension for MPS simulations."""
+    TRUNCATION: float = 1e-12
+    """The error threshold for dynamically truncating the bond dimension of MPS."""
+
+
+qatmpsconfig = QatMPSConfig()
+
```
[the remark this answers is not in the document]
lcauser-oqc: Done, let me know if this is what you had in mind.

Had to move the `QatMPSConfig` class to avoid circular imports. Could put it in a new file, but seemed simpler to me to just add it to `QatConfig`.

src/qat/purr/backends/qiskit_simulator.py
```diff
@@ -35,6 +36,12 @@ def get_default_qiskit_hardware(
     strict_placement=True,
     connectivity: Optional[Union[Connectivity, List[Tuple[int, int]]]] = None,
 ) -> "QiskitHardwareModel":
+    """
+    Creates a hardware model compatible with the Qiskit simulator.
+
+    If `strict_placement=True`, circuits can only be executed when circuit
+    intructions can be mapped directly onto the coupling map.
```
hamidelmaazouz, line 43: a bit unclear, "directly" as in "no swaps added or routing performed" ?

src/qat/purr/backends/qiskit_simulator.py
```diff
@@ -223,7 +230,34 @@ def __init__(self, hardware_model: QiskitHardwareModel = None):
...
+            "method": options.pop("method", "automatic"),
+            "matrix_product_state_max_bond_dimension": options.pop(
+                "matrix_product_state_max_bond_dimension",
+                qatconfig.MPS.MAX_BOND_DIMENSION,
+            ),
+            "matrix_product_state_truncation_threshold": options.pop(
+                "matrix_product_state_truncation_threshold", qatconfig.MPS.TRUNCATION
+            ),
+        }
+        config.update(options)
```
hamidelmaazouz, line 248: minor: but you've already done the dict merge manually above (sort of) ...

src/qat/purr/backends/qiskit_simulator.py
```diff
@@ -247,14 +281,40 @@ def execute(self, builder: QiskitBuilder) -> Dict[str, Any]:
             AerSimulator._DEFAULT_CONFIGURATION | {"open_pulse": False}
         )
         aer_config.n_qubits = self.model.qubit_count
-        qasm_sim = AerSimulator(aer_config, noise_model=builder.model.noise_model)
+
+        # AerSimulator can call a range of backends. We allow these to be choosen, along with
+        # any relevant arguments.
+        config = self.create_config(kwargs)
```
hamidelmaazouz, line 287: custom options could probably come from qat config ... some section that says: simulation -> backend -> method, ...

src/qat/purr/backends/qiskit_simulator.py
```diff
@@ -247,14 +281,40 @@ def execute(self, builder: QiskitBuilder) -> Dict[str, Any]:
...
         try:
             job = qasm_sim.run(
                 transpile(circuit, qasm_sim, coupling_map=coupling_map),
                 shots=builder.shot_count,
             )
             results = job.result()
+
+            # The job can fail if it runs into e.g. memory issues using the statevector
+            # simulator. In this case, and if no method is provided, try again using an
+            # MPS backend.
```
hamidelmaazouz, line 299: the optimistic retry could also be configured as a sequence in qat config .... The tail of which is the default or the fail back route

src/qat/purr/backends/qiskit_simulator.py
```diff
@@ -348,23 +410,35 @@ def execute(
...
+        https://docs.quantum.ibm.com/api/qiskit/0.37/qiskit.providers.aer.AerSimulator
+
+        Returns the measurements as a dict of bitstrings and associated counts.
+        If `return_metadata = true`, the metadata for the backend simulator is also
+        returned, i.e., returns as results, metadata = runtime.execute()
+        """
         if not isinstance(builder, QiskitBuilder):
             raise ValueError("Wrong builder type passed to QASM runtime.")
-
         # TODO - add error_mitigation for Qasm sim
```
hamidelmaazouz, line 427: rogue non applicable TODO, could you remove while you're at it ?

hamidelmaazouz (changes requested):

src/qat/purr/backends/qiskit_simulator.py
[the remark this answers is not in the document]
bgsach: Looks good

bgsach (approved):

keriksson-rosenqvist (approved):

src/qat/purr/backends/qiskit_simulator.py
lcauser-oqc: Done: for now, just done simulation -> method. Happy to revisit further branching if this becomes more extensive :smile:

src/qat/purr/qatconfig.py
```diff
@@ -1,13 +1,36 @@
...
+from pydantic import BaseModel, ConfigDict, Field, field_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
+from qiskit_aer import AerSimulator
 
 from qat.purr.utils.logger import get_default_logger
 
 log = get_default_logger()
 
 
+class QatSimulationConfig(BaseModel):
```
hamidelmaazouz, line 12: Question: How generic and reusable this config specification is ?
I mean can I use this config maybe in the future to specify simulations against a backend other than AerSimulator?
If no then that means there needs to be semantic cushioning between this config and how AerSimulator was actually specified.
Might not be the case, qatconfig is part of the public API, we'd need to avoid annoying renames, config repurposes, which would only lead to deprecations and headaches.
If it's only specific for Qiskit, then please rename to `QiskitSimulationConfig`

hamidelmaazouz (approved):
A few cosmetic remarks but all looks good. Thanks !

src/qat/purr/qatconfig.py
lcauser-oqc: Good point. While the MPS configuration would be reusable, I certainly wouldn't want to call the arguments by their Qiskit names. Lots of the other parameters are specific to the many implementations AerSimulator can call. Let's go with `QiskitSimulationConfig`. Thanks!

PR #208, opened 2024-10-04
Completely decouple package allocation from installation

hamidelmaazouz:
- Allows Q1 assembly debugging before installation
- Decouples Qblox resource allocation from package installation. This makes the install() method easily reusable for other internal purposes.


bgsach:
As we are moving to automated changelogs from PRs,
could you pop a general audience text description on this please and add the word Qblox to the title.

bgsach (approved):
LGTM

PR #209, opened 2024-10-07
Bump to 2.2.0

daria-oqc:


bgsach (changes requested):
Hmm, I'm hesitant to release a version which depends on released compiler_config

bgsach:
> Hmm, I'm hesitant to release a version which depends on released compiler_config

Fixed

bgsach (approved):

PR #210, opened 2024-10-09
[Qblox] Reuse complex-shaped waveforms whenever possible

hamidelmaazouz:
- When the waveform envelope is not trivial, try to find an existing waveform entry with the same data.
- Allows further optimisation on the waveform memory


bgsach:
Could you pop Qblox in the title of Qblox only PRs please so it's clearer in the release notes (in the future, I think we should adopt GitHub PR labels but this a quick stop gap)

bgsach (approved):
Looks good to me.

daria-oqc (approved):

PR #214, opened 2024-10-10
Bugfix: capture in openpulse

lcauser-oqc:
Using `capture` in openpulse will generate an acquire instruction that has a delay `None`, which gives errors when trying to do arithmetic with `NoneType` and `float`

Changes:
- If it exists, use the default delay for a measure acquire.
- Added a redundancy measure to ensure that if the delay is `None`, use `0.0` instead.
- Test to ensure it takes the default value.


src/qat/purr/integrations/qasm.py
```diff
@@ -1995,12 +2001,36 @@ def _capture_iq_value(
         # these post processing operations are removed for executions on live hardware.
         #
         # The returned value for each shot after postprocessing is a complex iq value.
+
+        # Determine the delay for the channel
+        delay = 0.0
+        if pulse_channel.channel_type == ChannelType.acquire:
+            # TODO: remove full id when previous PR is merged
```
bgsach, line 2008: Is this TODO something you can action now?

lcauser-oqc: Will action in #202 :smile:

src/qat/purr/backends/qblox/live.py
```diff
@@ -69,7 +69,9 @@ def _execute_on_hardware(
                 for aq in aqs:
                     physical_channel = aq.physical_channel
                     dt = physical_channel.sample_time
-                    physical_channel.readout_start = aq.start * dt + aq.delay
+                    physical_channel.readout_start = aq.start * dt + (
+                        aq.delay if aq.delay else 0.0
```
keriksson-rosenqvist, line 73: Minor but I would prefer more explicit if checks.

keriksson-rosenqvist (approved):

src/qat/purr/backends/qblox/live.py
bgsach: like `aq.delay or 0.0` ?

bgsach: (the last comment was a joke in case that wasn't clear)

keriksson-rosenqvist: :joy:

PR #220, opened 2024-10-14
Fix post-processing missing from map and add test.

keriksson-rosenqvist:
Fixing a bug where `PostProcess` instructions were not properly mapped into channels in `engine.create_duration_timeline`. Also added a test to verify the generated timeline has the expected types.

### DIFF
--- a/tests/qat/test_quantum_backend.py
+++ b/tests/qat/test_quantum_backend.py
@@ -30,9 +30,13 @@ from qat.purr.compiler.devices import (
 from qat.purr.compiler.emitter import InstructionEmitter, QatFile
 from qat.purr.compiler.hardware_models import QuantumHardwareModel
 from qat.purr.compiler.instructions import (
+    Acquire,
     Delay,
+    DrivePulse,
+    MeasurePulse,
     PhaseReset,
     PhaseShift,
+    PostProcessing,
     SweepValue,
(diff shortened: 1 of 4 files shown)


tests/qat/test_quantum_backend.py
```diff
@@ -677,3 +681,37 @@ def test_empty_channels_removed(self):
...
             )
+
+    def test_create_duration_timeline_mapping(self):
+        """Test that instructions are correctly mapped to expected channels."""
+        hw = get_test_model()
+        qubit = hw.get_qubit(0)
+
+        builder = (
+            get_builder(hw).X(qubit, np.pi / 2.0).measure_mean_z(qubit).synchronize(qubit)
+        )
```
hamidelmaazouz, line 692: can you parametrise the builder ? I'd ideally play with different combinations of instructions and see what happens

hamidelmaazouz (approved):

tests/qat/test_quantum_backend.py
keriksson-rosenqvist: We could, however that would also affect the assertions. Potentially better to add separate tests in the instances you'd like to test.

daria-oqc (approved):
Looks good to me, thanks for the patch!

PR #221, opened 2024-10-14
 Make OpenPulseFeatures python 3.11+ compatible

bgsach:
This fixes a bug where the OpenPulseFeatures class (which generates a json_dict describing our openpulse support on a given hardware model) was not python 3.11 compatible. This incompatibility was caused by one of it's supporting dataclasses having a default which was another (mutable) dataclass. This is not supported but was not checked until python 3.11.

Fortunately, the class(es) in question are (we believe) immutable in practice and so we have marked them immutable with the frozen=True attribute.

This was only spotted downstream because this file was not covered by tests (it wasn't even imported). I have added a basic test which fails without the fix.

### DIFF
--- a/src/qat/purr/integrations/features.py
+++ b/src/qat/purr/integrations/features.py
@@ -54,7 +54,7 @@ class Unit(Enum):
 
 # TODO: You don't want something lke this, just have two different objects for
 #   time/frequency.
-@dataclass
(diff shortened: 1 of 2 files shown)


daria-oqc (approved):
Thanks!

lcauser-oqc (approved):
Looks good!

PR #222, opened 2024-10-15
Extract append measure block to private method.

keriksson-rosenqvist:
Extract the code to append a measure block to a private method that accepts `**kwargs`.


lcauser-oqc (approved):
LGTM :smile:

daria-oqc (approved):

PR #223, opened 2024-10-16
Allow acquire time to be a Variable.

keriksson-rosenqvist:
Adjust acquire time validation to allow `Variable`.

### DIFF
--- a/src/qat/purr/compiler/instructions.py
+++ b/src/qat/purr/compiler/instructions.py
@@ -345,7 +345,11 @@ class Acquire(QuantumComponent, QuantumInstruction):
         super().__init__(channel.full_id())
         super(QuantumComponent, self).__init__(channel)
         self.time: float = time or 1.0e-6
-        if self.time < 0:
+        if not isinstance(self.time, Variable | float):
+            raise TypeError(
+                f"Acquire time must be of type 'float' or 'Variable', got {type(self.time)}."
+            )
+        if isinstance(self.time, float) and self.time < 0:
             raise ValueError(
                 f"Acquire time {self.time} cannot be less than or equal to zero."
             )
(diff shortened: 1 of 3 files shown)


tests/qat/test_instructions.py
```diff
@@ -318,6 +319,29 @@ def test_deviceinjector_reverts(self):
             execute_instructions(EchoEngine(hw), builder)
         assert measure_channel.frequency == 9e9
 
+    def test_sweep_acquire_time(self):
+        hw = get_default_echo_hardware(2)
+        qubit = hw.get_qubit(0)
+        measure_channel = qubit.get_measure_channel()
+        acquire_channel = qubit.get_acquire_channel()
+        acquire_times = [1.0e-6, 2.0e-6, 3.0e-6]
```
daria-oqc, line 327: If I understand correctly, the acquire times in the `Variable` can still be <0?

src/qat/purr/compiler/instructions.py
```diff
@@ -345,7 +345,11 @@ def __init__(
         super().__init__(channel.full_id())
         super(QuantumComponent, self).__init__(channel)
         self.time: float = time or 1.0e-6
-        if self.time < 0:
+        if not isinstance(self.time, Variable | float):
+            raise TypeError(
+                f"Acquire time must be of type 'float' or 'Variable', got {type(self.time)}."
+            )
+        if isinstance(self.time, float) and self.time < 0:
```
daria-oqc, line 352: Makes sense, thanks!

daria-oqc (approved):

daria-oqc:
Approved under the condition that my comment above gets addressed.

tests/qat/test_instructions.py
daria-oqc: EDIT: After a discussion with @keriksson-rosenqvist , we decided to ignore the comment above as it will cause a lot of breaks in dependent packages. 

We will consider this comment in the HWModel refactor.

bgsach (approved):

hamidelmaazouz (approved):
Could be a

lcauser-oqc (approved):
LGTM! :smile:

PR #226, opened 2024-10-17
Improvement/ker/phase reset drop

keriksson-rosenqvist:
Clean up instructions added by `builder.measure()`. This no longer adds a `PhaseReset` at the end and only syncs the measured `Qubit`s' channels at the start, not all entangled devices.


keriksson-rosenqvist:
@bryn-bell Here is the PR I mentioned. Let me know how you want to proceed.

src/qat/purr/compiler/instructions.py
```diff
@@ -869,9 +869,7 @@ def instructions(self) -> List[QuantumInstruction]:
         instructions = [Synchronize(list(self._entangled_qubits))]
```
willh727-OQC, line 869: Is it also worth changing this to targets? It seems to be part of the confusion

willh727-OQC (approved):
We should also think about stripping back the complex of measure, it seems to be producing a lot of unexpected behavior

hamidelmaazouz (approved):

src/qat/purr/compiler/instructions.py
keriksson-rosenqvist: This will expand to all `Qubit`s measured in the `MeasureBlock`. Does not care about entanglement.

willh727-OQC: I ment measurement targets not the entangled_qubits! So the conclusions is to change this to measurement targets from chats.

keriksson-rosenqvist: Done as of last commit.

lcauser-oqc (approved):
LGTM!

PR #227, opened 2024-10-17
Bugfix: sech waveforms

lcauser-oqc:
Sech waveforms throw a value error due to `mpmath`, which I suspect was used instead of `numpy` to avoid overflow errors...

Changes:
- Use numpy to calculate cosh
- Truncation of the input data to avoid overflow errors
- Some testing of sech waveforms


src/qat/purr/backends/utilities.py
```diff
@@ -204,12 +205,24 @@ def eval(self, x: np.ndarray) -> np.ndarray:
...
     def eval(self, x: np.ndarray) -> np.ndarray:
-        return 1 / cosh(x / self.width)
+        # Having a narrow width can cause overflows in numpy
+        # Restricting the argument such that cosh is within the max float range
+        # will overcome this, and has a neglibable effect (as sech(x) outside this
+        # range is practically zero).
+        x = np.array(
+            [max(min(np.real(val) / self.width, MAX_COSH_ARG), -MAX_COSH_ARG) for val in x]
+        )
+        return 1 / np.cosh(x)
```
bgsach, line 225: How about this

`np.clip(x.real / width, -MAX_COSH_ARG, MAX_COSH_ARG)`

Timed in a notebook, It's much faster

bgsach (changes requested):

bgsach (approved):

daria-oqc (approved):
LGTM

PR #229, opened 2024-10-23
Performance regression checks: benchmarking tests

lcauser-oqc:
**Performance regression checks when merging to main.**

This first PR adds a number of performance benchmarks that will enable us to ensure that we do not introduce performance regressions to QAT.

- Added `pytest-benchmark` to allow us to run performance benchmarks.
- Introduced a variety of benchmarks in the form of QASM files that will test compilation performance (let me know if there are any more you would like to see...).
- Updated `README` to explain how the benchmarks can be used.

The workflow to automatically compare performance to main follows in #235.

Note: There are some temporary benchmark tests that are only there for demonstration purposes, and will be removed before merging.


.github/workflows/benchmarking-regression.yml
```diff
@@ -0,0 +1,72 @@
...
+        with:
+          path: .venv
+          key: venv-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}-${{ hashFiles('**/poetry.lock') }}
+        
+      - uses: actions/checkout@v4
+        with:
+          ref: main
+
+      - name: Install dependancies
+        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
```
bgsach, line 52: I think this if is actually broken. There is no step cached-poetry-dependencies.outputs and It looks like it will always be true.

Given that .venv is a cache, poetry install --sync is fast so it's pretty harmless to run it.

.github/workflows/benchmarking-regression.yml
```diff
@@ -0,0 +1,72 @@
...
+        run: poetry install --sync
+      
+      - name: Install pytest-benchmark
+        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
+        run: poetry run pip install pytest-benchmark
```
bgsach, line 57: why do we install this seperately? it would be cleaner to do poetry install --sync --with dev on the previous step.

.github/workflows/benchmarking-regression.yml
bgsach: I've made a ticket for this as it's a bug inherited from the package-build.yaml

.github/workflows/benchmarking-regression.yml
```diff
@@ -0,0 +1,72 @@
...
+        run: poetry run pytest --benchmark-only --benchmark-save="benchmark"
+
+      - uses: actions/checkout@v4
+        with:
+          clean: false
+
+      - name: Install dependancies
+        if: steps.cached-poetry-dependencies.outputs.cache-hit != 'true'
+        run: 
+          poetry install --sync
```
bgsach, line 69: This if introduces a bug because we don't reload the cache the cache-hit is against the main poetry.lock.

Dropping the if statement seems fine or alternatively reloading the cache.

.github/workflows/benchmarking-regression.yml
lcauser-oqc: This is just a temporary measure (as pytest-benchmark) isn't currently a dependency on main. Turns out it was for nothing anyway as workflow fails due to a benchmarking file not being generated on main (because no benchmarks currently exist) :sweat_smile:

But yes this should be removed before being merged :smile:

README.rst
```diff
@@ -104,6 +104,18 @@ Additionally, to report any concerns or
 
 |
 
+----------------------
+Benchmarking
+----------------------
+
+The performance of QAT can be measured using :code:`poetry run pytest --benchmark-only`, which runs our pre-defined benchmarks.
+To compare to main, checkout the main branch and run :code:`poetry run pytest --benchmark-only --benchmark-save="benchmark"`.
```
keriksson-rosenqvist, line 112: To avoid confusion it would be good to use the _standard_ variable notation here: `--benchmark-save="<benchmark-name."` unless it it explicitly the string `benchmark` that needs to be provided.

Also would be good to have a link to the docs for further reading.

keriksson-rosenqvist (approved):
LGTM

bgsach (approved):

PR #232, opened 2024-10-24
Improved measurement instruction sequence

keriksson-rosenqvist:
Simplifies and improves the instructions that are generates by the `builder.measure()` method.

- Removed the `PhaseReset` at the end of the sequence, allowing continued operations
- `Synchronize` only across the measured targets, not all _possibly_ entangled targets

As seen in the test changes this means many circuits will become shorter.


lcauser-oqc (approved):

daria-oqc (approved):

tests/qat/test_quantum_backend.py
```diff
@@ -703,16 +703,13 @@ def test_create_duration_timeline_mapping(self):
         assert [type(pos.instruction) for pos in drive_data] == [
             DrivePulse,
             Delay,
-            PhaseReset,
```
hamidelmaazouz, line 706: even for drive channel ?

keriksson-rosenqvist: Yes, that's the requested behaviour.

PR #234, opened 2024-10-25
Temporary stop releases with version mismatch.

keriksson-rosenqvist:
Temporarily block/reject releasing and publishing docs based on tags if the version tag does not match both the package version and the version referenced in the docs config.
@alillistone-OQC Would like you're input on this solution.


bgsach (approved):
I manually tested the core `$(grep -oP 'version = "\K.*?(?=")' docs/source/conf.py)` line locally.

alillistone-OQC (approved):
Tested the reg exp locally and can confirm it works. This looks good from me

PR #236, opened 2024-10-28
Pull out the new pass-style QuantumRuntime to new module path.

keriksson-rosenqvist:
Move the new pass-infrastructure `QuantumRuntime` to a new module `qat/runtime/runtime.py` to comply with the new package structure which is used for the next generation QAT compiler and runtime.

### DIFF
--- a/src/qat/purr/compiler/runtime.py
+++ b/src/qat/purr/compiler/runtime.py
@@ -13,10 +13,6 @@ from compiler_config.config import (
     ResultsFormatting,
 )
 
-from qat.compiler.transform_passes import PhaseOptimisation, PostProcessingOptimisation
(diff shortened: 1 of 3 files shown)


lcauser-oqc (approved):
LGTM

hamidelmaazouz (commented):
I would not do that unless there's a blocking reason.

Part of the refactoring we're drafting is to depart away from the hand-me-down cascade like stages: Frontend, Runtime, Engine, Instrument, .... I do not want to see the new QAT structure reproduce the same structure.

`NewQuantumRuntime` is only temporarily here to serve Qblox hw sweeps. It is under `purr` because it still adheres to the legacy structure. The new `qat.runtime` will have a completely different outlook.

keriksson-rosenqvist:
Closing based on Hamid's input

PR #237, opened 2024-10-31
Resolve np.tanh warnings from soft square waveforms

lcauser-oqc:
Fix the overflow errors thrown by `np.tanh(x)` warnings in various waveforms. The error is thrown (on windows) when the argument `x` is complex with a sufficiently large real component. However, the waveforms are only evaluated against time (a real number), so treating `x` as complex is unnecessary. 

Changes:
- Call `x.real` but return the results as a complex type for consistency (there is no restriction that the waveform returned cannot be complex).
- Added some tests for affected waveforms.


keriksson-rosenqvist (approved):
LGTM assuming the complex maths are as expected.

daria-oqc (approved):
Thanks for picking this up!

PR #239, opened 2024-11-04
Performance improvement: mid-circuit measurement validation

lcauser-oqc:
**Context:**  validating that an instruction list does not have mid-circuit measurements is currently very expensive. It checks that there are no mid circuit measurements by iterating through the entire instruction set, and checking if a qubit has an acquire instruction appear before a pulse. To do so, it has to resolve the qubit for each pulse. However, this only needs to practically be done once per pulse channel, so we can cache the result and reuse that.

**Changes:**

- Cached the result of resolving_qb_pulse_channel for each quantum target so that it can be reused.
- Fixed an error in the validation that compared acquire and measure channels (maybe this needs to be more robust to consider *both* acquires and measures)
- Added a test to ensure mid-circuit measurements are caught.

Results in an order of magnitude improvement for validate for deep circuits (likely to be less impactful for shallow circuits)

### DIFF
--- a/src/qat/compiler/validation_passes.py
+++ b/src/qat/compiler/validation_passes.py
@@ -129,7 +129,7 @@ class ReadoutValidation(ValidationPass):
             # Check if we've got a measure in the middle of the circuit somewhere.
             elif isinstance(inst, Acquire):
                 for qbit in model.qubits:
-                    if qbit.get_measure_channel() == inst.channel:
+                    if qbit.get_acquire_channel() == inst.channel:
                         consumed_qubits.append(qbit)
             elif isinstance(inst, Pulse):
                 # Find target qubit from instruction and check whether it's been
--- a/src/qat/compiler/validation_passes.py
+++ b/src/qat/compiler/validation_passes.py
@@ -84,7 +84,7 @@ class InstructionValidation(ValidationPass):
 
 class ReadoutValidation(ValidationPass):
     """
-    Extracted from LiveDeviceEngine.optimize()
+    Extracted from LiveDeviceEngine.validate()
     """
 
     def run(self, builder: InstructionBuilder, res_mgr: ResultManager, *args, **kwargs):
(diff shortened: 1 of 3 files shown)


src/qat/compiler/validation_passes.py
```diff
@@ -128,13 +129,23 @@ def run(self, builder: InstructionBuilder, res_mgr: ResultManager, *args, **kwar
             # Check if we've got a measure in the middle of the circuit somewhere.
             elif isinstance(inst, Acquire):
                 for qbit in model.qubits:
-                    if qbit.get_measure_channel() == inst.channel:
```
daria-oqc, line 131: Is there any reason this was previously implemented as `get_measure_channel` instead of `get_acquire_channel`?

daria-oqc (approved):
Thanks!

src/qat/compiler/validation_passes.py
lcauser-oqc: I'm not too sure if this was a mistake or intentional but I'll try to find out. Thinking the check might need to be more robust (e.g. look for acquires measures)

tests/qat/test_quantum_backend.py
```diff
@@ -716,3 +716,23 @@ def test_create_duration_timeline_mapping(self):
...
+        engine = get_test_execution_engine(hw)
+        q0 = hw.get_qubit(0)
+        q1 = hw.get_qubit(1)
+        builder = (
+            get_builder(hw)
+            .had(q0)
+            .X(q1)
+            .measure(q0)
+            .measure(q1)
+            .cnot(q0, q1)
```
keriksson-rosenqvist, line 733: Super minor but cloud we also add tests to validate:
1. measure only one of the qubits in the cnot before
2. measure e.g. q2 and then do gates on un-measured qubits

keriksson-rosenqvist (approved):
LGTM, though would like the additional tests for completeness.

tests/qat/test_quantum_backend.py
lcauser-oqc: Thanks, new tests coming right up! :smile:

PR #240, opened 2024-11-05
[EXPERIMENTAL] Serialisation of QuantumHardwareModels

daria-oqc:
This PR introduces the new hardware model that is serialised using [Pydantic](https://docs.pydantic.dev/latest/why/). The current version of `QuantumHardwareModel` has the following devices
- physical basebands
- physical channels
- pulse channels
- qubits
- resonators

which are each mapped by their unique ID. These fields are immutable, which in practice means that we cannot just add/delete qubits etc. after a hardware model has been instantiated. We can, however, change the frequency of a pulse channel or the drive amplitude of a qubit after the hardware model has been instantiated.

The hardware model can be serialised/deserialised using the following pseudocode (kudos to @bgsach)
```
hw = QuantumHardwareModel(qubits=..., resonators=..., pulse_channels=..., physical_channels=..., physical_basebands=...)

serialised_hw = hw.model_dump()

deserialised_hw = QuantumHardwareModel(**serialised_hw)

hw == deserialised_hw
>>> True
```

Additionally, I have gathered as much information as possible on the variables that each device can hold (e.g., `frequency`, `if_frequency`, `scale`, etc.). Please feel free to add any additional details if you think the current documentation could be improved.

I will follow up with other PRs that implement the hardware model builder, and incorporate the new hardware model into our current compilation workflow.


src/qat/model/ref.py
bgsach, line 1: Rename this annotations.py now it also contains the IdDict?

daria-oqc: As discussed through private channels, we are renaming it to `serialisation.py`.

src/qat/model/hardware_model.py
```diff
@@ -0,0 +1,48 @@
...
+        resonators: The resonators on the chip.
+    """
+
+    version: SemanticVersion = VERSION
+    physical_basebands: ComponentDict[PhysicalBaseband] = Field(frozen=True, default=dict())
+    physical_channels: ComponentDict[PhysicalChannel] = Field(frozen=True, default=dict())
+    pulse_channels: ComponentDict[PulseChannel] = Field(frozen=True, default=dict())
+    qubits: ComponentDict[Qubit] = Field(frozen=True, default=dict())
+    resonators: ComponentDict[Resonator] = Field(frozen=True, default=dict())
+
```
lcauser-oqc, line 39: There's a whole bunch of "helper" functions in the legacy hardware models used to retrieve certain properties (e.g. finding pulse channels from physical channels). Is the plan to design other parts of QAT such that these are no longer needed? Or will these come later?

src/qat/model/serialisation.py
```diff
@@ -0,0 +1,79 @@
...
+Ref = Annotated[ComponentId | T, SerializeItemToRefs, "Ref"]
+
+SerializeDictToRefs = PlainSerializer(
+    lambda d: {k.uuid: v.model_dump_id() for (k, v) in d.items()},
+    return_type=dict,
+    when_used="always",
+)
+
+
+def deserializeDictToRefs(data):
```
lcauser-oqc, line 24: Super minor point but isn't the standard in python to have function names with underscores instead of caps? e.g. `deserialize_dict_to_refs(data)`. I personally don't mind as long as it is consistent :)

Also typo below "Reydrates"

src/qat/model/device.py
```diff
@@ -0,0 +1,117 @@
...
+        default_pulse_channel_type: Default type of pulse for the qubit.
+    """
+
+    measure_device: Ref[Resonator] = Field(frozen=True)
+
+    index: int = Field(ge=0)
+    drive_amp: float = 1.0
+    default_pulse_channel_type: ChannelType = Field(frozen=True, default=ChannelType.drive)
+
+    coupled_qubits: Optional[RefList[Qubit]] = None
```
lcauser-oqc, line 117: Do we have any plans for how to deal with the pulse property dicts in legacy Qubits? i.e. `pulse_hw_x_pi_2`,  `pulse_measure` and `measure_acquire `

tests/qat/model/test_pydantic_hw_serialisation.py
```diff
@@ -0,0 +1,134 @@
+from __future__ import annotations
```
lcauser-oqc, line 1: Could the file names for tests mimic the structure of `src/qat/` (e.g. as unit tests so we can easily identify where tests should go)

lcauser-oqc (commented):
Really great work Daria! This already looks so much cleaner and simpler than the legacy code :smile: 

I have some remarks below (mostly questions to help me better understand the bigger picture :sweat_smile:)

src/qat/model/device.py
```diff
@@ -0,0 +1,117 @@
...
+        frequency: Frequency of the pulse.
+        bias: Mean value of the signal, quantifies offset relative to zero mean.
+        scale: Scale factor for mapping the voltage of the pulse to frequencies.
+        fixed_if: Flag which determines if the intermediate frequency is fixed.
+        channel_type: Type of the pulse.
+        auxiliary_qubits: Any extra devices this PulseChannel could be affecting except
+                           the current one. For example in cross resonance pulses.
+    """
+
+    physical_channel: Ref[PhysicalChannel]
```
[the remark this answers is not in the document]
daria-oqc: Agree with this. Ideally, the physical channel should only "live" in the quantum device but since the current code uses the pulse channels to access the physical channel, I am not sure this is something we can easily introduce without refactoring a lot of code.

src/qat/model/device.py
daria-oqc: Yes that will be implemented in the next stage.

keriksson-rosenqvist (approved):
LGTM
Do we want to also have a test for deserialising a versioned file?

bgsach (approved):

daria-oqc:
> LGTM Do we want to also have a test for deserialising a versioned file?

@keriksson-rosenqvist This is implicitly taken care of in [this line](https://github.com/oqc-community/qat/blob/d3328f10341e4583aa8cd1e54283e3ae222609e0/tests/qat/model/test_serialisation_hardware_model.py#L107), where the version numbers of the serialised and deserialised hardware models must match. However, I have added an extra explicit test as well for the version number (de)serialisation.

PR #241, opened 2024-11-05
Minor performance improvement: get_X_from_Y

lcauser-oqc:
Extracted from #202 

**Context:** We frequently call methods that find one component of a hardware model from another, e.g., finding a device from a pulse channel. When doing so, we provide the full_id of the component X, and the function will then look into the hardware model and retrieve X, and then use this to find Y. However, quite often we already have X, meaning we pass the argument `get_X_from_Y(X.full_id())`.

**Changes**:

- Allow the arguments for these functions to take either a `str` or `type(X)`.
- Change these function calls to use `X` instead of `X.full_id()`


bgsach (approved):

keriksson-rosenqvist (approved):
LGTM as long as we're sure it doesn't cause issues with serialisation.

daria-oqc (approved):

PR #244, opened 2024-11-06
[EXPERIMENTAL] Builder for `QuantumHardwareModel`

daria-oqc:
- [x] Completely refactored the `QuantumHardwareModel`. It is now much simpler to serialise and deserialise since the hardware model can now be represented by a graph with a tree structure according to: `Qubit` -> `Resonator` -> `PulseChannels` -> `PulseChannel`, `Qubit` -> `PulseChannels` -> `PulseChannel` and `Qubit` -> `PhysicalChannel` -> `PhysicalBaseband`. This means we can safely omit `autopopulate.py` and `serialisation.py`.
- [x] Provided a hardware model builder that requires the topology of the qubits in the QPU. Since not all qubit couplings may be of the same quality, we can constrain the topology of the hardware model to only include the couplings we want.
- [x] Tests tests tests.


src/qat/model/device.py
```diff
@@ -1,12 +1,75 @@
...
+    Base class for any logical object which can act as a target of a quantum action
+    - a Qubit or various channels for a simple example.
+
+    Attributes:
+        uuid: The unique string representation of the component.
+    """
+
+    model_config = ConfigDict(validate_assignment=True, extra="ignore")
+
+    uuid: str = Field(default_factory=lambda: str(uuid.uuid4()), frozen=True)
```
daria-oqc, line 25: Still not sure if we should keep the `uuid` or not. This question will probably answer itself when we merge the functionality of these hardware models with the new pipeline stuff.

src/qat/model/device.py
```diff
@@ -55,64 +117,111 @@ class PulseChannel(Component):
...
+    def __ne__(self, other: PulseChannels):
+        return not self.__eq__(other)
+
+
+class ResonatorPulseChannels(PulseChannels):
+    measure: MeasurePulseChannel
+    acquire: AcquirePulseChannel
+
+
+class QbloxResonatorPulseChannels(PulseChannels):
```
daria-oqc, line 191: In a later stage, we can move this class to the `qblox` directory.

tests/qat/model/test_device.py
```diff
@@ -0,0 +1,2 @@
+class TestDevicesValidation:
+    pass
```
daria-oqc, line 2: I'll add some appropriate tests here once we all agree with the proposed structure of the `QuantumHardwareModel`.

src/qat/model/hardware_model.py
```diff
@@ -1,48 +1,124 @@
...
+        for node, connected_nodes in topology.items():
+            for connected_node in connected_nodes:
+                if node not in topology[connected_node]:
+                    raise ValueError(
+                        f"The topology is not symmetric, node {node} not present in connected nodes of node {connected_node}."
+                    )
+        return topology
+
+    @field_validator("constrained_topology")
+    def validate_constrained_topology_symmetry(cls, constrained_topology):
```
lcauser-oqc, line 93: Does the topology actually need to be symmetric here?

For example, a particular hardware model might directly support a CNOT with control qubit 4 and target qubit 5, but not (directly) support a CNOT with control qubit 5 and target qubit 4. Of course in practice we can always indirectly achieve the CNOT with the opposite direction by use of Hadamard gates.

My thoughts are if this wasn't required to be symmetric, this could, for example, guide the compilation in determining how to implement the CNOT in the opposite direction (for this example).

lcauser-oqc (approved):
This is great, so much simpler without the `RefDicts` etc!!!

src/qat/model/device.py
lcauser-oqc: I think you're right, keep it until we know we don't need it (although because of the hashing, I expect that this will still be useful!)

src/qat/model/hardware_model.py
```diff
@@ -1,48 +1,124 @@
...
+
+class QuantumHardwareModel(LogicalHardwareModel):
+    """Class for calibrating our QPU hardware.
+
+    Attributes:
+        qubits: The superconducting qubits on the chip.
+        constrained_topology: The connectivities between the qubits, which can be a subgraph of the physical topology.
+    """
+
+    qubits: dict[QubitId, Qubit]
```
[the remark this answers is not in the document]
daria-oqc: We _could_. However, I believe such solution would cause some overhead if we decide to remove some qubits from the hardware model due to them malfunctioning. In this case, there is no clear 1:1 mapping between the qubit index and the list index anymore..

src/qat/model/hardware_model.py
daria-oqc: I honestly have no idea! You could be right, I'll ask around to give you a more definite answer on this.

src/qat/model/hardware_model.py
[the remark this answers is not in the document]
lcauser-oqc: Ah thanks, that makes sense!

keriksson-rosenqvist (commented):
Great work!
:tada: :clap: 

Could we add tests for the main known topologies as well, i.e. Lucy and Toshiko?

src/qat/model/device.py
```diff
@@ -1,118 +1,248 @@
...
+                return False
+        return True
 
 
 class PhysicalBaseband(Component):
     """
     Models the Local Oscillator (LO) used with a mixer to change the
     frequency of a carrier signal.
+
     Attributes:
```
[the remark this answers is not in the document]
bgsach: I much prefer the numpy style but am not suggesting we change now :)

src/qat/model/hardware_model.py
bgsach: OK, I think we need to remove this symmetric validator.

src/qat/model/hardware_model.py
```diff
@@ -1,48 +1,124 @@
...
+    def __ne__(self, other: LogicalHardwareModel) -> bool:
+        return not self.__eq__(other)
+
+
+class QuantumHardwareModel(LogicalHardwareModel):
+    """Class for calibrating our QPU hardware.
+
+    Attributes:
+        qubits: The superconducting qubits on the chip.
+        constrained_topology: The connectivities between the qubits, which can be a subgraph of the physical topology.
```
bgsach, line 59: I'm not sure about the word 'constrained' but I can't think of a good alternative.

More generally, I was imagining that the roles of the two topologies were reversed:

physical_topology on the physical hardware model logical_topology on the logical hardware model

with logical_topology required to be a subset of the physical_topology but only physical models (logical only models can have whatever topology they like)

src/qat/model/device.py
```diff
@@ -1,118 +1,248 @@
...
         frequency: The frequency of the LO.
         if_frequency: The intermediate frequency (IF) resulting from
                       mixing the baseband with the carrier signal.
     """
 
-    frequency: float = Field(ge=0.0)
-    if_frequency: Optional[float] = Field(ge=0.0, default=250e6)
+    frequency: float = Field(default=np.nan)
```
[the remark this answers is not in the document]
daria-oqc: Yes since we aim to use `np.nan` for an uncalibrated hw parameter, and `np.nan` will not pass `> 0` during validation

src/qat/model/device.py
```diff
@@ -1,118 +1,248 @@
...
-    baseband: Ref[PhysicalBaseband] = Field(frozen=True)
+    baseband: PhysicalBaseband = Field(frozen=True)
 
-    sample_time: float = Field(ge=0.0)
+    sample_time: float = Field(default=np.nan)
     block_size: Optional[int] = Field(ge=1, default=1)
-    phase_iq_offset: float = 0.0
-    bias: float = 1.0
-    acquire_allowed: bool = False
+    phase_iq_offset: complex = 0.0 + 0.0j
```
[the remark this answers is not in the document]
daria-oqc: We _can_, do you see any advantages in using the numpy version over the internal version?

src/qat/model/device.py
```diff
@@ -1,118 +1,248 @@
...
+    frequency: float = Field(default=np.nan)
     bias: complex = 0.0 + 0.0j
     scale: complex = 1.0 + 0.0j
     fixed_if: bool = False
 
-    channel_type: Optional[ChannelType] = Field(frozen=True, default=None)
-    auxiliary_qubits: RefList[Qubit] = []
 
+class MeasurePulseChannel(PulseChannel):
+    pass
```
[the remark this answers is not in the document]
daria-oqc: I'll change it to the ellipsis format as this is what is the standard in QAT, I guess.

src/qat/model/device.py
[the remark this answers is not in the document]
bgsach: [https://docs.pydantic.dev/2.3/api/types/#pydantic.types.confloat - allow_inf_nan](https://docs.pydantic.dev/2.3/api/fields/#:~:text=_Unset-,allow_inf_nan,_Unset,-max_digits)

src/qat/model/device.py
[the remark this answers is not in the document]
daria-oqc: I have adjusted it to the Sphinx style, without the types of the params as these are more than clear in the model field type specifications.

daria-oqc:
> Could we add tests for the main known topologies as well, i.e. Lucy and Toshiko?
@keriksson-rosenqvist Thanks for this really insightful comment, I've added some tests for both architectures. Let me know if you have remarks/other suggestions.

src/qat/model/builder.py
hamidelmaazouz, line 1: all apis here seem to stress over the "uncalibrated" nature of the to-be-built hw model, a few clarifying questions:
* how important is the specification of this concept here ?
* do you think it's a concern of the builder here ?

I have the impression usage intent leak into these apis. It looks as if you're giving hints as to where they can or can't be called. But calibration is a whole different concern, and you want to keep apis here as impartial as you can ...

src/qat/model/device.py
```diff
@@ -1,118 +1,272 @@
...
+import numpy as np
+from pydantic import Field, model_validator
+from pydantic_core import core_schema
 
-from qat.model.component import Component
-from qat.model.serialisation import Ref, RefDict, RefList
-from qat.purr.compiler.devices import ChannelType
+from qat.utils.pydantic import WarnOnExtraFieldsModel
+
+QubitId = int
```
hamidelmaazouz, line 12: minor: can be hoisted up and put it somewhere common to all hw model stuff, to void specifying types all over ...

src/qat/model/device.py
```diff
@@ -1,118 +1,272 @@
...
+    def __get_pydantic_core_schema__(cls, source_type, handler) -> core_schema.CoreSchema:
+        return core_schema.no_info_plain_validator_function(cls.validate)
+
+    def __eq__(self, other: CalibratablePositiveFloat):
+        if np.isnan(self) and np.isnan(other):
+            return True
+        else:
+            return self.__eq__(other)
+
+    def __neq__(self, other: CalibratablePositiveFloat):
```
hamidelmaazouz, line 32: does not add much tbh, it's just a an inverse of `__eq__()` call

src/qat/model/device.py
```diff
@@ -1,118 +1,272 @@
...
+    uuid: str = Field(default_factory=lambda: str(uuid.uuid4()), frozen=True)
+
+    def __init__(self, **data):
+        super().__init__(**data)
+
+    def __hash__(self):
+        return hash(self.uuid)
+
+    def __eq__(self, other: Component):
+        if type(self) != type(other):
```
hamidelmaazouz, line 53: eh ?! do I understand you want them to have the exact same type ?

src/qat/model/device.py
```diff
@@ -1,118 +1,272 @@
...
+    """
+    Base class for any logical object which can act as a target of a quantum action
+    - a Qubit or various channels for a simple example.
+
+    :param uuid: The unique string representation of the component.
+    """
+
+    uuid: str = Field(default_factory=lambda: str(uuid.uuid4()), frozen=True)
+
+    def __init__(self, **data):
```
hamidelmaazouz, line 46: trivial constructor, unless you (plan to) have attributes to define (uuid ??) then it could be removed

src/qat/model/device.py
hamidelmaazouz: maybe you meant `isinstance()` ?

src/qat/model/device.py
```diff
@@ -1,118 +1,272 @@
...
+        super().__init__(**data)
+
+    def __hash__(self):
+        return hash(self.uuid)
+
+    def __eq__(self, other: Component):
+        if type(self) != type(other):
+            return False
+
+        if self.model_fields != other.model_fields:
```
hamidelmaazouz, line 56: this sounds like comparing between two collections, fine if they are sets but I'd be skeptical in general

src/qat/model/device.py
```diff
@@ -1,118 +1,272 @@
...
+            return False
+
+        if self.uuid != other.uuid:
```
hamidelmaazouz, line 59: you probably want to start with this check first before you check sth heavier

src/qat/model/device.py
```diff
@@ -1,118 +1,272 @@
...
+        return not self.__eq__(other)
+
+    def __repr__(self):
+        return f"{self.__class__.__name__}({self.uuid})"
+
+    def __str__(self):
+        return self.__repr__()
+
+    @property
+    def calibrated(self):
```
hamidelmaazouz, line 84: `is_calibrated()`

src/qat/model/device.py
```diff
@@ -1,118 +1,272 @@
...
 
-from qat.model.component import Component
-from qat.model.serialisation import Ref, RefDict, RefList
-from qat.purr.compiler.devices import ChannelType
+from qat.utils.pydantic import WarnOnExtraFieldsModel
+
+QubitId = int
+
+
+class CalibratablePositiveFloat(float):
```
hamidelmaazouz, line 15: there are two important ideas just from the name:

1. Tunability of a float number
2. The float remaining positive during the tuning process

Could explain why do you need the second point ?

src/qat/model/device.py
hamidelmaazouz, line 1: Design here needs a rethink, I see a mix of many different concepts:

* Component definition and identification
* Specification of components' properties
* Backend leaks like Qblox

I also see a tendency to cramp everything in one python module. Start by distinguishing:

* Components of the cryo system: Qubit, Resonator, ...
* Components of the the control stack: PysicalChannel, PhysicalBaseband, ...

These are the concepts, then define:

1. what the apis they must fullfil to any caller
2. their identification system
3. flexibility of customers defining perperties with customised terminology

The three points above are to be treated separately per component kind

hamidelmaazouz (changes requested):

src/qat/model/device.py
daria-oqc: `model_fields` are the names of the member variables in the given `Component`, so we check if they contain the same attributes

src/qat/model/device.py
bgsach: Let's chat about this

hamidelmaazouz (commented):
Unblocking in case the team wants to go ahead with this design as a first iteration

src/qat/model/builder.py
daria-oqc: I added the `uncalibrated` hint as the parameters in the hardware model, such as the pulse channel frequencies, will all be set to `np.nan`. As such, the user needs to take care of supplying appropriate values for these parameters.

src/qat/model/device.py
daria-oqc: This variable is used for frequencies and sample times, which are by definition never lower than zero (unless I'm horribly mistaken?). So we validate against the positivity condition

PR #245, opened 2024-11-08
[EXPERIMENTAL] Improvements to binding and legalisation passes

hamidelmaazouz:
* `BindingPass` builds binding of variables, instructions, and a view of variables from/to scopes.
* (Target-agnostic) `LegalisationPass` performs target-independent legalisation. Currently on legalise how iterator variables are used and legalise their bounds.


lcauser-oqc (approved):

src/qat/backend/analysis_passes.py
```diff
@@ -179,85 +177,196 @@ def extract_iter_bound(name: str, value: Union[List, np.ndarray]):
...
+                    }
+        if stack:
+            raise ValueError(f"Unbalanced scopes. Found orphans {stack}")
 
-            for name, grp in groups.items():
-                if len(grp) > 1:
+        res_mgr.add(result)
+
+
+class TILegalisationPass(AnalysisPass):
```
willh727-OQC, line 284: Could we avoid acronyms here! Also is target-independent really that clear? It seems its doing SweepLegalisation no?

PR #247, opened 2024-11-13
Added error_mitigation as a property for QuantumExecutionEngine

lcauser-oqc:
**Context:** Sometimes hardware models and execution engines are used interchangeably. When using error readout mitigation, this can cause errors when validating the model against error mitigation due to the fact `error_mitigation` is not a property of `QuantumExecutionEngine`.

**Changes**:
- Added `error_mitigation` as a property of `QuantumExecutionEngine`.
- Added tests to ensure validation doesn't raise an error.


daria-oqc (approved):
Thanks!

keriksson-rosenqvist (approved):
LGTM

PR #255, opened 2024-11-15
Improved test coverage for default QASM gates.

keriksson-rosenqvist:
Tests that verify expected QASM 2 and 3 gates are working. Clearly lists exceptions for gates that are not (yet) supported.


bgsach (approved):

src/qat/purr/backends/echo.py
```diff
@@ -76,8 +76,8 @@ def apply_setup_to_hardware(
                         frequency=5.5e9,
                         scale=0.0,
                     )
-                qubit.add_coupled_qubit(qubit_devices[(i + 1) % qubit_count])
-                qubit.add_coupled_qubit(qubit_devices[(i - 1) % qubit_count])
+            qubit.add_coupled_qubit(qubit_devices[(i + 1) % qubit_count])
+            qubit.add_coupled_qubit(qubit_devices[(i - 1) % qubit_count])
```
hamidelmaazouz, line 80: can't believe it, was it adding to a set or a list ?

src/qat/purr/integrations/qasm.py
```diff
@@ -983,6 +983,8 @@ def visit_QuantumGate(self, node: ast.QuantumGate, context: QasmContext):
             case "u" | "U":
                 theta, phi, _lambda = arguments
                 return self.add_unitary(theta, phi, _lambda, target_qubits, self.builder)
+            case "cu" | "CU":
+                raise NotImplementedError("Controlled Unitary gate is not yet supported.")
```
hamidelmaazouz, line 987: Do we have an equivalent ticket/issue ?

tests/qat/qasm_utils.py
```diff
@@ -71,6 +71,13 @@ def parse_and_apply_optimiziations(
     return builder
 
 
+qasm3_base = """OPENQASM 3.0;
+bit[{N}] c;
+qubit[{N}] q;
+{gate_strings}
+measure q -> c;"""
```
hamidelmaazouz, line 78: super minor and more of personal preference:
```
qasm3_base = """
OPENQASM 3.0;
bit[{N}] c;
qubit[{N}] q;
{gate_strings}
measure q -> c;
"""
```

src/qat/purr/backends/echo.py
keriksson-rosenqvist: The `Qubit` has a method that actually does nothing if it already has that coupling, but we were still going into that method and checking it an equivalent coupling already exists way more times than we needed. :loop:

PR #256, opened 2024-11-19
[EXPERIMENTAL] Skeptical static `DeviceUpdate` instructions

hamidelmaazouz:
A static `DeviceUpdate` is one that updates an attribute on a target with a value that's known at compile time.

For example `DeviceUpdate(some_channel, "scale", 1)` simply assigns `1` to the attribute `scale` on target `some_channel`. But how come people use this version instead of simply saying `some_channel.scale = 1` ?

In fact, the legacy code processes `DeviceUpdate`s with a structure called `DeviceInjector` not only to "inject" (i.e. assign at runtime) but to "revert" any previous assignment(s) at "some aftermath" (*). It's the injection + reversion guarantee that appeals to people more than the direct way. Therefore `DeviceUpdate(some_channel, "scale", 1)` is roughly equivalent to sth like:

```
try:
    old_scale = some_channel.scale
    some_channel.scale = 1
finally:
    some_channel.scale = old_scale
```

or via some mechanism based on Python contexts.

Legacy code w/r to `SweepIterator` and `DeviceInjectors` worked fine because the (QAT IR) programs were simple basic blocks. In light of the recent efforts to bring control flow, analysis of these programs become (semantically) intractable. Therefore the problem being raised here is more on what static `DeviceUpdate` instructions mean globally for the QAT IR program and in awareness of control flow. Here are a few claims:

1. While they are perfectly valid instructions, people use static `DeviceUpdate`s (only) to benefit from the assignment + reversion process described above (*).
2. Programmers writing QAT IR are **likely unaware** that the position of a static `DeviceUpdate` among other instructions is ultimately important.
3. Static `DeviceUpdate`s can ultimately influence codegen. Pulse evaluation at compile time is one typical effect.

This PR only describes a peculiar case of `DeviceUpdate` in QAT IR and does not provide a definite solution. The workaround implemented in this PR assumes the claims laid out above. Here's what it contributes:

* Adapt existing analyses to recognise static `DeviceUpdate`s
* Filter out any static `DeviceUpdate`s prior to codegen
* Process static `DeviceUpdate`s globally and irrespective of any control flow considerations

### DIFF
--- a/src/qat/backend/analysis_passes.py
+++ b/src/qat/backend/analysis_passes.py
@@ -72,7 +72,7 @@ class TriagePass(AnalysisPass):
                 else:
                     targets.update(inst.quantum_targets)
 
-            if isinstance(inst, DeviceUpdate):
+            if isinstance(inst, DeviceUpdate) and isinstance(inst.value, Variable):
                 reads[inst.target].add(inst.value.name)
 
         result = TriageResult()
@@ -303,7 +303,7 @@ class BindingPass(AnalysisPass):
                         ]
                 elif isinstance(inst, Acquire):
                     rw_result.writes[inst.output_variable].append(inst)
-                elif isinstance(inst, DeviceUpdate):
+                elif isinstance(inst, DeviceUpdate) and isinstance(inst.value, Variable):
                     if not (
                         inst.value.name in scoping_result.symbol2scopes
                         and [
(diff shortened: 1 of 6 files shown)


lcauser-oqc (approved):

daria-oqc (approved):
LGTM

PR #257, opened 2024-11-19
Patch modified gates with QASM2 implementations.

keriksson-rosenqvist:
Add an internal version of `stdgates.inc` with patched gate definitions.

PR #258, opened 2024-11-19
Patch modified gates with QASM2 implementations.

keriksson-rosenqvist:
Twin to https://github.com/oqc-community/qat/pull/257


lcauser-oqc (approved):

lcauser-oqc (approved):

daria-oqc (approved):

src/qat/purr/integrations/grammars/stdgates.inc
hamidelmaazouz, line 1: I assume this was copied from https://github.com/Qiskit/qiskit/blob/main/qiskit/qasm/libs/stdgates. ?

src/qat/purr/integrations/qasm.py
```diff
@@ -816,10 +816,12 @@ def _walk_program(self, prog: ast.Program, context: QasmContext):
             self.visit(node, context)
 
     def visit_Include(self, node: ast.Include, context: QasmContext):
-        if node.filename in ("stdgates.inc", "qelib1.inc"):
+        if node.filename == "qelib1.inc":
             file_path = Path(
                 find_spec("qiskit.qasm.libs").submodule_search_locations[0], node.filename
             )
+        elif node.filename == "stdgates.inc":
```
hamidelmaazouz, line 823: can't we just reuse `find_spec` to locate the same file ?
oh or is the `"stdgates.inc"` file not available in the qiskit release we depend on ?

src/qat/purr/integrations/qasm.py
```diff
@@ -983,24 +985,30 @@ def visit_QuantumGate(self, node: ast.QuantumGate, context: QasmContext):
...
                 return self.add_ecr(target_qubits, self.builder)
 
+        if len(node.modifiers) > 0:
+            raise NotImplementedError("Gate modifiers are not yet supported.")
+
         gate_context = QasmContext(
             Registers(),
             context.gates,
-            dict(context.variables),
+            dict(),
```
hamidelmaazouz, line 999: intuitively speaking, this is a fresh clean context, not sure if this is intentional, and whether a gate would not need access to prior defined variables

hamidelmaazouz (approved):
What are the json files for ?

src/qat/purr/integrations/grammars/stdgates.inc
keriksson-rosenqvist: Initially, then overloaded with definitions we support

src/qat/purr/integrations/qasm.py
keriksson-rosenqvist: Because of the way gate definitions look, they all use the same variables for the input qubits which causes issues when the definition needs to "redefine" what qubit is referred to.

PR #259, opened 2024-11-20
Test coverage: gate decomposition tests

lcauser-oqc:
**Context**: We currently allow -- either via our `QuantumInstructionBuilder` or parsing `QASM2/3` files -- many different gates, that are ultimately implemented through decomposition down to the native gates typically available on hardware. While we have some tests that indirectly check some of these gates through simulators (e.g. by inspecting measurement distributions), they are not exhaustive and robust to the subtle details that might distinguish gates (for example, the `Rx` and `Ry` gate applied to a `|0>` state would give the same measurement distribution, but there are phase differences that wont be detected). Having more robust tests will allow us to move forwards with more trust in our internal and external definitions of gates. The most complete way to compare decompositions of gates against their standard definitions would be to compare unitary matrices of decomposed circuits against their expected outcomes.

These tests also detected that our implementation of the Hadamard gate at the level of the builder is incorrect (this is not a problem when running QASM programs due to the its overloaded definition).

**Changes**:

- Added an augmented hardware model and instruction builder to our testing package to allow us describe circuits as unitary matrices.
- These are used to test decompositions of gates supported in our `QuantumInstructionBuilder` against standard definitions.
- Also tests decompositions of gates from the `qelib1.inc` library in QASM2.
- Tests for QASM3 with the updated `stdgates.inc` library.
- Hadamard patch

Tests currently fail as we need the QASM3 patch on main first :smile:


src/qat/purr/compiler/builders.py
```diff
@@ -285,8 +285,8 @@ def swap(self, target: Qubit, destination: Qubit):
         raise NotImplementedError("Not available on this hardware model.")
 
     def had(self, qubit: Qubit):
-        self.Y(qubit, math.pi / 2)
-        return self.Z(qubit)
+        self.Z(qubit)
+        return self.Y(qubit, math.pi / 2)
```
daria-oqc, line 289: 👍🏻

tests/qat/test_gates.py
```diff
@@ -0,0 +1,238 @@
...
+    params = product(thetas, thetas, thetas)
+    for args in params:
+        func_to_gates[f"U({args[0]}, {args[1]}, {args[2]})"] = ["U", Gates.U(*args), args]
+
+    return func_to_gates
+
+
+class TestSingleGates:
+
+    model = get_default_matrix_hardware(1)
```
daria-oqc, line 46: Just to verify it's still working on multi-qubit systems, can we also have tests for `n_qubits > 1`?

daria-oqc (approved):

tests/qat/test_gates.py
lcauser-oqc: So I somewhat confusingly placed a test that does this for `n_qubits = 2` under `TestDoubleGates`, but this could definitely be more robust (e.g. I'm only doing this for single qubit gates, we could also verify `n_qubits > 2` for two-qubit gates) :sweat_smile: 

I'll add some more, thanks! :smile:

keriksson-rosenqvist (approved):
LGTM

PR #262, opened 2024-11-22
[EXPERIMENTAL] Finetune hardware model according to internal discussions.

daria-oqc:


src/qat/model/hardware_model.py
```diff
@@ -58,30 +62,66 @@ class PhysicalHardwareModel(LogicalHardwareModel):
...
                     "Logical connectivity must be a subgraph of the physical connectivity."
                 )
 
-        return data
+        return self
+
+    @model_validator(mode="after")
+    def validate_connectivity_quality(self):
+        for q1_index, q2_index in self.physical_connectivity_quality:
+            if not self.physical_connectivity.get(q1_index, None):
```
daria-oqc, line 112: I assume that the quality of the physical connectivity/coupling between the qubits is directional as well? I.e., `quality(qubit1, qubit2) != quality(qubit2, qubit1)`?

Can someone confirm this?

lcauser-oqc: I agree with this! 

Out of curiosity, what decides the value of the quality? Is it just the measured fidelity of the 2q gate?

daria-oqc: cc/ @bryn-bell @elskysr

src/qat/model/hardware_base.py
```diff
@@ -1,3 +1,68 @@
...
+    float,
+    Predicate(validate_calibratable_unit_interval),
+    validate_calibratable_unit_interval,
+]
+
+K = TypeVar("K")
+V = TypeVar("V", CalibratablePositiveFloat, CalibratableUnitInterval)
+
+
+class ValidatedDict(RootModel[dict[K, V]]):
```
daria-oqc, line 44: Sadly, pydantic does not validate its mutable fields (such as dictionaries) after instantiation. See a similar discussion [here](https://stackoverflow.com/questions/75396856/pydantic-does-not-validate-the-key-values-of-dict-fields). This is why we need to give these mutable fields an extra layer of validation. This is currently a prototype which works for `physical_connectivity_quality` but should be general enough to be applied for all dictionaries within the hardware model.

src/qat/model/hardware_base.py
```diff
@@ -1,3 +1,215 @@
...
+import numpy as np
+from frozendict import frozendict
+from pydantic import AfterValidator, RootModel
+from pydantic_core import core_schema
+
 # This base file is used to implement classes/methods common to all hardware.
 
-QubitId = int
+
+def validate_non_negative(v: int):
```
keriksson-rosenqvist, line 14: Minor but I'd prefer `val` to v, even if it's fairly simple functionality it still is nice to be explicit.

src/qat/model/hardware_base.py
```diff
@@ -1,3 +1,215 @@
...
+    return v
+
+
+CalibratableUnitInterval = Annotated[
+    float,
+    AfterValidator(validate_calibratable_unit_interval),
+]
+
+
+K = TypeVar("K")
```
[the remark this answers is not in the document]
daria-oqc: `K` and `V` are just a general type for the keys and values in dicts, or just values in sets. do you prefer other names?

src/qat/model/hardware_base.py
daria-oqc: I'll change it, thanks!

PR #263, opened 2024-11-22
Fix poetry cache id

alillistone-oqc:
The conditional for the poetry install step was using a step id that didn't exist. This will correct this

### DIFF
--- a/.github/workflows/benchmarking-regression.yml
+++ b/.github/workflows/benchmarking-regression.yml
@@ -49,6 +49,7 @@ jobs:
           ref: main
 
       - name: Install dependancies
+        if: steps.cached-pip-wheels.outputs.cache-hit != 'true'
         run: poetry install --sync
 
       - name: Benchmarking on main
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmarking-regression.yml
```diff
@@ -49,6 +49,7 @@ jobs:
           ref: main
 
       - name: Install dependancies
+        if: steps.cached-pip-wheels.outputs.cache-hit != 'true'
```
hamidelmaazouz, line 52: anything but `true` is False ?

hamidelmaazouz (approved):

.github/workflows/benchmarking-regression.yml
alillistone-OQC: Yeah if cache-hit is true (makes the expression false) then we do not want to install dependencies because they have been imported in by a prior step. The step is skipped. If cache-hit is false (makes this expression true) then install dependencies.

PR #267, opened 2024-11-25
[EXPERIMENTAL] Pydantic instructions for faster serialization

lcauser-oqc:
Adapting legacy QAT's instructions to be derived from Pydantic's `BaseModel` so we can use Pydantic for faster (de)serialization. Some key points are:

- The "targets" of instructions, typically PulseChannels, are stored within the instruction in legacy QAT. This entangling of instructions and hardware models makes serialization messy and slow (as we have to implement our custom JSON pickle method which is slow). Replacing this targets with their IDs helps in this regard, but means we have to pay a price elsewhere in the compilation where we would have to retrieve the targets from the model.
- To make use of Pydantic's serialization, we have to add a "type" member to each instruction so the correct instruction can be retrieved on deserialization. Tried out different approaches, but this gave the best performance improvements.
- This work involves bringing `instructions.py` and `waveforms.py` to experimental. It's worth emphasising that this will probably look very different in later iterations of new QAT, but for now is necessary to bring this down so we can test and map out this approach. Currently placed in `qat/ir`, but this might be wrong...
- Some reorgnaization of instructions, including bringing out measurement related instructions to its own package, and a refactor of pulses/waveforms.
- A first-pass minimal working example will have the QAT pipeline include a `serialize` and `deserialize` function that converts between legacy and Pydantic instructions, and serializes/deserializes at the level of Pydantic. The long-term plan would be to slowly introduce Pydantic instructions to other areas of new QAT as they are more developed, but for now this gives us a way to test the new instructions without having to rework them into all other areas of the code base.

Some rough benchmarks (with the details to be shared elsewhere):
- When working solely with Pydantic Instructions, we see performance improvements in serializaiton in the range of 25-70x on a few select examples. Similarly, deserialization is in the range to 4-10x faster.
- When working with legacy instructions, and using the conversion mechanisms to serialize using pydantic, we see speed-ups of ~5-20x for serialization and ~2x-4x in deserialization.

Currently includes other files purely for testing and measuring performance.


src/qat/core.py
```diff
@@ -6,15 +6,20 @@
...
+from qat.purr.compiler.builders import InstructionBuilder, QuantumInstructionBuilder
 from qat.purr.compiler.frontends import QASMFrontend
 from qat.purr.compiler.hardware_models import QuantumHardwareModel
 from qat.purr.compiler.metrics import CompilationMetrics
 from qat.purr.qatconfig import QatConfig
 from qat.qat import QATInput
+from qat.utils.instruction_conversion import (
+    legacy_to_pydantic_instructions,
+    pydantic_to_legacy_instructions,
+)
```
hamidelmaazouz, line 22: delegate this to a `IRConverter` object

src/qat/core.py
```diff
@@ -46,6 +51,31 @@ def execute(
...
+        The blob must have been generated using a Pydantic InstructionList. However,
+        the deserialized instructions are return as the legacy instructions inside a
+        QuantumInstructionBuilder. This process makes use of the hardware model to
+        repopulate the instructions with the necessary PulseChannels.
+        """
+        pydantic_instructions = InstructionList.deserialize(blob)
+        instructions = pydantic_to_legacy_instructions(
+            pydantic_instructions, self._hardware_model
+        )
+        return QuantumInstructionBuilder(self._hardware_model, instructions)
```
hamidelmaazouz, line 77: nope, delegate to a deserialiser/streamser object

src/qat/ir/emitter.py
hamidelmaazouz, line 1: needs a redesign, these are codegen concepts

src/qat/ir/emitter.py
```diff
@@ -0,0 +1,180 @@
...
+    def __repr__(self):
+        sep = ";" if any(self.reliant) else ""
+        sched_deps = f"[{','.join(self.dependencies)}]" if any(self.dependencies) else ""
+        return (
+            f"{self.instruction.duration}: {sched_deps} {str(self.instruction)}"
+            f"{sep}{';'.join(str(inst) for inst in self.reliant)}"
+        )
+
+
+class QatFile(BaseModel):
```
hamidelmaazouz, line 69: Can do without, use the `TriagePass`

src/qat/ir/emitter.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        Takes a JSON blob and returns a QAT file.
+        """
+        blob = from_json(blob)
+        return QatFile(**blob)
+
+
+class InstructionEmitter:
+    """
+    Interface to the LLVM-driven QAT optimization and construction passes. For now we
```
hamidelmaazouz, line 139: imprecise comment, yes it's inspired by LLVM, but it's not driven by llvm ^^, we're just borrowing a few concepts and techniques ^^

src/qat/ir/instructions.py
hamidelmaazouz, line 1: needs a lot more work, it's an opportunity to systematise QAT semantics and reason better about iterations, variables, acquisitions, and posprocessing headaches.

advice: try to depart away from what instructions looked in legacy code and come up with completely new structures

src/qat/ir/emitter.py
lcauser-oqc: Yeah I'm going to remove this, I don't think we need to bring the emitter down. The point is to get something minimal we can use with the legacy code, which I think is serializing before this point (e.g. after instructions are parsed through the builder).

src/qat/ir/emitter.py
lcauser-oqc: Haha yeah I agree, this comment was from the legacy code :sweat_smile:

src/qat/ir/instructions.py
```diff
@@ -0,0 +1,1032 @@
...
+    States that this value is actually a variable that should be fetched instead.
+    """
+
+    name: str
+    # TODO: we might consider limiting what types a variable can take...
+    var_type: Optional[type] = None
+    value: Any = None
+    model_config = ConfigDict(validate_assignment=True)
+
+    def __init__(self, name, var_type=None, value=None, **kwargs):
```
keriksson-rosenqvist, line 70: Do we need the `**kwargs` here?

src/qat/ir/instructions.py
```diff
@@ -0,0 +1,1032 @@
...
+        # find the correct type.
+        if isinstance(var_type, str):
+            var_type = locate(var_type)
+        return var_type
+
+    @field_validator("value", mode="after")
+    @classmethod
+    def _validate_value_type(cls, value, val_info):
+        var_type = val_info.data["var_type"]
+        if var_type != None and not isinstance(value, var_type):
```
keriksson-rosenqvist, line 102: How does this work if `value=None`? I.e. if I declare a `Variable("freq", var_type=float)` To be a frequency variable who's value should strictly be a float when set, but without setting it initially, e.g. to be used in a `Sweep`. :thinking:

src/qat/ir/instructions.py
```diff
@@ -0,0 +1,1032 @@
...
+
+
+class Return(Instruction):
+    """A statement defining what to return from a quantum execution."""
+
+    inst: Literal["Return"] = "Return"
+    variables: Optional[List[str]]
+
+    def __init__(self, variables: List[str] = None, **kwargs):
+        if variables is None:
```
keriksson-rosenqvist, line 184: Is there any need for this checking when you have the validator?

src/qat/ir/instructions.py
keriksson-rosenqvist, line 1: Generally on the instructions. What's the benefit of having all the `def __init__(...)` methods?
There are still a bunch of "legacy" input validation and manipulation happening. Why not replace all of that with the Pydantic validators?

src/qat/ir/instructions.py
```diff
@@ -0,0 +1,1032 @@
...
+
+    def __init__(self, target: QuantumComponent, attribute: str, value, **kwargs):
+        super().__init__(target, attribute=attribute, value=value)
+
+    def __repr__(self):
+        return f"{self.target}.{self.attribute} = {str(self.value)}"
+
+
+### Instruction blocks
+def _all_instructions(start_types=[Instruction]):
```
keriksson-rosenqvist, line 856: Redefining of a function in `instruction_list.py`. Even if it's a fairly short function there's risk of inconsistencies creeping in.

src/qat/ir/instructions.py
```diff
@@ -0,0 +1,1032 @@
...
+
+
+class InstructionBlock(BaseModel):
+    """
+    An Instruction grouping type. Allows working with blocks of Instructions as a unit.
+    """
+
+    inst: Literal["InstructionBlock"] = "InstructionBlock"
+
+    def _validate_types(self, items, valid_types, label="targets"):
```
keriksson-rosenqvist, line 886: This should probably be replaced by pydantic validators

src/qat/ir/instructions.py
```diff
@@ -0,0 +1,1032 @@
...
+
+    inst: Literal["MeasureBlock"] = "MeasureBlock"
+    target_dict: Dict[str, MeasureData] = {}
+    existing_names: Optional[Set[str]] = set()
+    duration: float = 0.0
+
+    @property
+    def instructions(self):
+        instructions = [Synchronize(self.pulse_channel_targets)]
+        for _, val in self.target_dict.items():
```
keriksson-rosenqvist, line 921: `self.target_dict.values()`

tests/qat/test_core.py
```diff
@@ -113,6 +122,25 @@ def equivalent_vars(self, other):
...
+        vars_other = vars(other)
+        if len(vars_self) != len(vars_other):
+            return False
+        for key in vars_self.keys():
+            if not key in vars_other:
+                return False
+            check = vars_self[key] == vars_other[key]
+            if not isinstance(check, bool):
+                check = all(check)
+                print(vars_self[key])
```
keriksson-rosenqvist, line 137: Prints?

tests/qat/test_core.py
```diff
@@ -113,6 +122,25 @@ def equivalent_vars(self, other):
...
+                print(vars_other[key])
+            if not check:
+                return False
+    return True
```
keriksson-rosenqvist, line 141: This seems to have the wrong indentation. `equivalent_array("Hello", 42)` looks to evaluate to `True`. :thinking:

keriksson-rosenqvist (commented):
A few comments from a rough scan.

src/qat/ir/instruction_list.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class InstructionList(BaseModel):
+    instruction_list: InstBlockList = []
+
+    @property
+    def instructions(self):
+        flat_list = []
+        for inst in self.instruction_list:
+            if isinstance(inst, MeasureBlock):
```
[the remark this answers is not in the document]
daria-oqc: From a conceptual point of view: is it possible to make the `Instruction` class a [composite](https://refactoring.guru/design-patterns/composite) that can have an arbitrary tree structure?

src/qat/ir/instructions.py
```diff
@@ -0,0 +1,1032 @@
...
+
+class Repeat(Instruction):
+    """
+    Global meta-instruction that applies to the entire list of instructions. Repeat
+    value of the current operations, also known as shots.
+    """
+
+    inst: Literal["Repeat"] = "Repeat"
+    repeat_count: Any
+    repetition_period: Any = None  # TODO: probably can only be a float, but not sure...
```
daria-oqc, line 133: Yes I would believe so

src/qat/ir/instructions.py
```diff
@@ -0,0 +1,1032 @@
...
+
+    def __init__(self, repeat_count: int = None, repetition_period=None, **kwargs):
```
daria-oqc, line 135: Is there a particular reason why you need this (trivial) `__init__`?

src/qat/ir/instructions.py
daria-oqc: Or just set the default value of `variables` to []?

src/qat/ir/instructions.py
```diff
@@ -0,0 +1,1032 @@
...
+    pass
+
+
+class SweepValue(SweepOperation):
+    name: str
+    value: Any
+
+
+### Variables
+def build_generated_name(existing_names=None, prefix=None):
```
daria-oqc, line 39: Why do we need to generate random names, can't we just use uuids?

src/qat/ir/instructions.py
```diff
@@ -0,0 +1,1032 @@
...
+
+
+class PhaseShift(QuantumInstruction):
+    """
+    A PhaseShift instruction is used to change the phase of waveforms sent down
+    the pulse channel.
+    """
+
+    inst: Literal["PhaseShift"] = "PhaseShift"
+    quantum_targets: str
```
daria-oqc, line 367: Why can't we make the type of this variable a `PulseChannel` or a collection thereof?

src/qat/ir/instructions.py
```diff
@@ -0,0 +1,1032 @@
...
+    phase: Union[float, Variable] = 0.0
+
+    def __init__(self, quantum_targets: PulseChannel, phase: float, **kwargs):
```
daria-oqc, line 370: See comment above on `__init__`

src/qat/ir/instructions.py
lcauser-oqc: Good point, added support for this!

tests/qat/test_core.py
lcauser-oqc: aha, remnants from debugging, thanks! :sweat_smile:

src/qat/ir/instructions.py
lcauser-oqc: Yeah would probably be easier (and safer) to use UUIDs. Not sure if there is anywhere else which is expected it to have a name with this structure though... Nevertheless, I haven't changed this from the legacy so I'll just change this to import for now

src/qat/ir/instructions.py
```diff
@@ -0,0 +1,1032 @@
...
+            raise ValueError(
+                f"channel has type {str(type(target))}: it must have type PulseChannel"
+            )
+        return target.full_id() if isinstance(target, PulseChannel) else target
+
+
+class Id(QuantumInstruction):
+    """Simply a no-op, called an Identity gate."""
+
+    # TODO: do we really need this?
```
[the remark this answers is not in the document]
lcauser-oqc: Sorry I'm not sure what you mean :sweat_smile: 

My comment just meant do we really need an instruction that does nothing? `Delay` tells a particular channel to do nothing for some time, this just seems like some mixing of "gates" and "instructions". I think it can probably go

src/qat/ir/instructions.py
lcauser-oqc: So with the legacy hardware models, since they're not derived on the `BaseModel`, this wouldn't work.

Of course with the new hardware models this isn't an issue. However, when we serialize, the details of the hardware model (e.g. pulse channels) would also be serialized. If we really wanted the instruction to store the entire pulse channel, we could probably do something with `field_serializers` and `field_validators` to deal with this.

My thoughts are it is best to separate this entirely (just storing ids/uuids). There is the obvious downside that we would need to retrieve the pulse channels at compilation. However, the `TriagePass` bundles instructions by pulse channels so this should only need to be done once per pulse channel.

src/qat/ir/instructions.py
```diff
@@ -0,0 +1,1032 @@
...
+    def __init__(
+        self,
+        quantum_targets: Union[Qubit, PulseChannel, List[Union[Qubit, PulseChannel]]],
+        **kwargs,
+    ):
+        super().__init__(quantum_targets=quantum_targets)
+
+    @field_validator("quantum_targets", mode="before")
+    @classmethod
+    def _pulse_channels_to_strs(cls, targets):
```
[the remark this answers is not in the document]
lcauser-oqc: See comment above

src/qat/ir/instructions.py
lcauser-oqc: So initially these were necessary to have the pydantic instructions working with legacy structures (e.g. a builder that directly makes pydantic instructions). The reason for this is mostly that legacy instructions use positional arguments, but also I have renamed some of the members for clarity.

However I moved away from this since I think there is agreement the builder needs a lot of work... I was questioning whether to keep these as there are some cases where positional arguments would no longer work if we removed them (removing this could also be a good thing to improve code clarity). But after seeing your's and @daria-oqc's comments, I think I agree its best to remove these (and move that complexity to the `IRConverter`)

src/qat/ir/instructions.py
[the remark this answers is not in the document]
daria-oqc: Oh I misunderstood the purpose of this class, thought it was a id tag. Anyway, it can be nice to have as a conceptual gate to improve the clarity for some circuits, I don't think it does any harm anyway

src/qat/ir/instructions.py
bgsach: Your focus here is right Luke, you aren't expected to fix everything in one PR. There are concrete improvements we need from the pydantic instructions.

PR #272, opened 2024-11-27
Adjust passes to use QatIR.

keriksson-rosenqvist:
Use `QatIR` type as input to all passes. Allows the same pipeline to operate from QASM/QIR all the way down to the executable instructions.


lcauser-oqc (approved):
LGTM

PR #292, opened 2025-11-18
qblox backend: configuration update for QCM

RJQblox:
This fix is needed since the NCO propagation delay is only available for readout modules (QRM), but not control modules (QCM). Both the `src/qat/backend/qblox` and `src/purr/cat/backends/qblox` folder have been updated, is this correct?

### DIFF
--- a/src/qat/backend/qblox/config/helpers.py
+++ b/src/qat/backend/qblox/config/helpers.py
@@ -41,13 +41,16 @@ class QbloxConfigHelper(ABC):
         if self.sequencer_config.connection.bulk_value:
             sequencer.connect_sequencer(*self.sequencer_config.connection.bulk_value)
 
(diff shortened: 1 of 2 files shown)


helmaazouz-oqc (changes requested):
Thank you @RJQblox for the suggested changes ! When it comes to configuration, QBlox have fine-grained APIs for numerous settings, we've grouped NCO related ones exactly to avoid fine-grained methods. But you still have several subtypes of `QBloxConfigHelper` for the purposes you mentioned, I encourage you to override nco there with the sequence of changes you like (excluding prop delay)

hamidelmaazouz:
@RJQblox 

Are there any follow ups on this PR ? do you still need assistance ? In general we're moving away from that style of configuration and I advise to follow up with those developments as they progress. I'll close the PR for now

